PR #3, opened 2018-07-08
[add] give possibility to configure default ursadb index types

raw-data:
- added `INDEX_TYPE` field in `config.example.py` 

- updated `lib/ursadb.py` accordingly to changes in `config.example.py` 

So that, added paths via the `mquery admin` page, can be indexed with custom index types as set in the config


icedevml:
Merged after small modifications. Thanks for your contribution.

PR #44, opened 2020-04-04
Expose ursadb taints in the frontend

msm-code:
Expose ursadb taints in the frontend.

This will allow users to query only specific datasets (for example, check their yara rule on known-good dataset).

This implementation is a bit limited compared to what the backend support (impossible to select multiple taints), but should be useful anyway.

### DIFF
--- a/src/lib/ursadb.py
+++ b/src/lib/ursadb.py
@@ -1,6 +1,10 @@
 import json
 import time
 import zmq  # type: ignore
+from typing import Dict, Any
+
+
+Json = Dict[str, Any]
 
 
 class UrsaDb(object):
@@ -42,7 +46,7 @@ class UrsaDb(object):
 
         return {"time": (end - start) * 1000, "files": files}
 
-    def status(self):
+    def status(self) -> Json:
         socket = self.make_socket()
         socket.send_string("status;")
         response = socket.recv_string()
@@ -50,7 +54,7 @@ class UrsaDb(object):
 
         return json.loads(response)
 
-    def topology(self):
+    def topology(self) -> Json:
         socket = self.make_socket()
         socket.send_string("topology;")
         response = socket.recv_string()
@@ -58,7 +62,7 @@ class UrsaDb(object):
 
         return json.loads(response)
 
-    def execute_command(self, command):
+    def execute_command(self, command: str) -> Json:
         socket = self.make_socket(recv_timeout=-1)
         socket.send_string(command)
         response = socket.recv_string()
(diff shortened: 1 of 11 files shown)


ITAYC0HEN:
would you mind adding screenshots and test plan? will test this tomorrow :)

msm-code:
Hi,

sure! I've actually done this a few days ago and assigned for review internally, but after discussion today decided to move most development to github - hence this PR.

Test plan:
 - checkout this commit
 - docker-compose build && docker-compose up
 - add some samples to ./samples directory, then `docker-compose run ursadb-cli tcp://ursadb:9281`  and:
```
> index "/mnt/samples";
OK
> topology;
+-----------+-----------+
| dataset   | indexes   |
|-----------+-----------|
| 738c8b6d  | gram3     |
+-----------+-----------+
> taint "738c8b6d" add "test";  # substitute "738c8b6d" for id from above
OK
```

Expected status page (frontend subject to change as soon as I figure out how to present it in a better way):
ss

Expected new button on the query page:

ss

src/lib/ursadb.py
```diff
@@ -15,11 +19,15 @@ def make_socket(self, recv_timeout=2000):
...
         return socket
 
-    def query(self, query):
+    def query(self, query, taint):
         socket = self.make_socket(recv_timeout=-1)
 
         start = time.clock()
-        query = "select {};".format(query)
+        if taint:
+            taint = taint.replace('"', '"')
```
nazywam, line 27: Is there a situation in which this isn't a NOP? If that's the case, perhaps a small comment explaining it?

src/lib/ursadb.py
```diff
@@ -15,11 +19,15 @@ def make_socket(self, recv_timeout=2000):
         socket.connect(self.backend)
         return socket
 
-    def query(self, query):
+    def query(self, query, taint):
```
nazywam, line 22: Since we're adding type hints in other functions of this class we might upgrade this one as well?

src/lib/ursadb.py
msm-code: :thinking: what the... Yeah, that is supposed to be: 

```python
taint = taint.replace('"', '\\"')
```

src/lib/ursadb.py
msm-code: Sure! I went ahead and added other missing type hints here

msm-code:
Merge (to avoid future merge conflicts) since looks like it's ready to go. Feedback still welcome though!

ITAYC0HEN:
sorry for not testing, still waiting for the indexing and don't want to touch it too much (again :( )

msm-code:
No worries, this is just a small change anyway - you'll have a lot to test later :).

PR #55, opened 2020-04-06
Allow users to specify multiple rules

### LINKED ISSUE #40: Support multiple rules and dependencies between them
**Description**

*Out of all the issues I opened, this probably is the most important to me.*

mquery currently supports the querying of only a single rule in a given query. That means that set of rules is not supported. This issue is important to me in particular since this is a usual research workflow, and in general, a very basic feature of Yara.

Not only that mquery does not support referencing rules in the same query, but it also does not support multiple rules querying at the time. And I mean multiple rules for the same query, and not for different queries running on multiple workers.

Most of the Yara rules repositories will contain `.yar` files with multiple rules inside. So are the rules that are written by different research vendors or by different individuals. A typical use case is a rule to catch all samples of a given actor, say a [ruleset to catch samples of Equation group](https://github.com/Neo23x0/signature-base/blob/master/yara/spy_equation_fiveeyes.yar). Mquery does not allow to do something like this, which is very common when working on big projects, analyzing entire actor and teams.

For me, the queries I usually create contain more than a single rule. It contains [global rules](https://yara.readthedocs.io/en/stable/writingrules.html#global-rules), [private rules](https://yara.readthedocs.io/en/stable/writingrules.html#private-rules) and referencing other rules. This allows me to create powerful rulesets with stronger conditions and classification. [In the end](https://www.youtube.com/watch?v=eVTXPUF4Oz4), the samples that will match the ruleset will be classified based on the rules in the ruleset that detected them. For example, in the aforementioned ruleset of Equation group, the results will show whether each sample detected by "EquationDrug_PlatformOrchestrator",  "Equation_Kaspersky_TripleFantasy_Loader", "Equation_Kaspersky_FannyWorm", or other rule.

Yara is very powerful, and it seems like mquery doesn't use its powers fully.

**Read more:**
[Yara - Referencing Other Rules](https://yara.readthedocs.io/en/stable/writingrules.html#referencing-other-rules)

```
rule Rule1
{
    strings:
        $a = "dummy1"

    condition:
        $a
}

rule Rule2
{
    strings:
        $a = "dummy2"

    condition:
        $a and Rule1
}
```

[Yara - More About Rules](https://yara.readthedocs.io/en/stable/writingrules.html#more-about-rules) - Global and Private rules

```
global rule SizeLimit
{
    condition:
        filesize < 2MB
}

private rule PrivateRuleExample
{
    ...
}
```

msm-code:
Another oversized PR from me. Sorry :thinking:

This will allow us to support querying using multiple yara
rules at once. Additionaly, elementary support of private and
global rules was added.

Test cases:

Trivial ones:

```
 rule Kot
{
    strings:
        $a = "xxx"

    condition:
        $a
}

rule Hmm
{
    strings:
        $a = "aaa"

    condition:
        $a and Kot
}
```

Equation group rules: https://raw.githubusercontent.com/Neo23x0/signature-base/master/yara/spy_equation_fiveeyes.yar

I think this fixes #40, but may need more refining in the future.

### DIFF
--- a/src/lib/yaraparse.py
+++ b/src/lib/yaraparse.py
@@ -1,5 +1,6 @@
-from yaramod import Yaramod  # type: ignore
-from yaramod import (
+import argparse
+from yaramod import (  # type: ignore
+    Yaramod,
     AndExpression,
     StringCountExpression,
     IntLiteralExpression,
@@ -11,11 +12,12 @@ from yaramod import (
     OfExpression,
     StringWildcardExpression,
     StringAtExpression,
+    IdExpression,
     StringInRangeExpression,
+    ThemExpression,
+    SetExpression,
 )
-from yaramod import ThemExpression, SetExpression
-import sys
-from typing import Optional, List
+from typing import Optional, List, Dict
 import re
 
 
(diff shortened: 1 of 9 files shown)


msm-code:
Screenshot: query results

ss

Screenshot: parse results

ss

src/mqueryfront/src/QueryField.js
```diff
@@ -164,7 +165,7 @@ class QueryField extends Component {
                             name="parse"
                             type="submit"
                             onClick={(event) =>
-                                this.handleQuery(event, "parse")
+                                this.handleQuery(event, "parse", "parse")
```
icedevml, line 168: :thinking: this task doesn't really have (any) priority, so why it is required?

msm-code: Hmm, long story. Yeah, not required in this case (TLDR earlier priority was in the path and `query/sth/undefined` looked ugly

src/lib/yaraparse.py
```diff
@@ -73,174 +145,206 @@ def ursify_string(string) -> Optional[str]:
...
+        else:
+            raise YaraParseError(f"Unsupported of_expr type: {type(children)}")
 
+        parsed_elements = [e for e in all_elements if e is not None]
 
-def of_expr(condition, rule_strings) -> Optional[str]:
-    how_many = condition.text[: condition.text.find("of")].strip()
-    counter = None
+        if how_many == "all":
+            counter = len(all_elements)
```
nazywam, line 229: I didn't notice this in the last review but shouldn't this be `len(parsed_elements)`?

Otherwise if we have `all of (x, y, z)` and `y` turns out to be a None in the ursa query, won't the final query look like `min 3 of (x, y)`?

src/lib/yaraparse.py
```diff
@@ -73,174 +145,206 @@ def ursify_string(string) -> Optional[str]:
...
 
-    assert len(rules.rules) == 1
+    for rule in rules:
+        if rule.is_global:
+            global_expressions.append(rule.parse())
+        elif not rule.is_private:
+            public_expressions.append(rule.parse())
 
-    rule = rules.rules[0]
+    print("public", public_expressions)
```
nazywam, line 321: :sad_print:

src/lib/yaraparse.py
msm-code: :thinking: that's tricky.

First of all, the change was intentional. But now I think it was not correct after all.

Let's consider three cases and what they should be simplified to:

```
all of ("aaa", "bbb", /regex/)  =>  (equivalent to) 2 of ("aaa", "bbb")?
```

```
3 of ("aaa", "bbb", /regex/)  =>  (equivalent to) 2 of ("aaa", "bbb")?
```

```
2 of ("aaa", "bbb", /regex/)  =>  (equivalent to) 1 of ("aaa", "bbb")?
```

In other words, I think using `all` as an alias for `len(all_elements)` here is correct, but I should also subtract number_of_none_elements from counter *in all cases*.

(ofc remember to cap counter at 0 from the bottom)

Please read my thoughts and tell me what you think :thinking:.

nazywam: So for `how_many == "all"` `len(all_elements) - number_of_none_elements` should be equal to `len(parsed_elements)` right?

But yes, subtracting it for all cases makes sense

PR #56, opened 2020-04-07
mwdb integration

KrzyWie:


src/metadata/MwdbGetMetadata.py
```diff
@@ -2,11 +2,11 @@
 
 import requests
 import urllib
-
+from . import config
```
msm-code, line 5: Is it used anywhere?

src/metadata/MwdbIntegrate.py
```diff
@@ -0,0 +1,54 @@
...
+
+        if not self.token:
+            logging.error(
+                "MWDB metadata not fetched, MWDB_API_TOKEN variable was not set"
+            )
+            return {}
+
+        try:
+            api = mwdblib.MalwarecageAPI(
+                api_url="https://mwdb.cert.pl/api/", api_key=self.token
```
msm-code, line 35: pass `api_url` from config (similarly to api_key)

src/metadata/MwdbIntegrate.py
```diff
@@ -0,0 +1,54 @@
...
+
+        except:
+            logging.exception("Failed to post mquery metadata")
+            return {}
+
+        obj = {}
+
+        obj["mwdb_sample"] = {
+            "display_text": "mwdb",
+            "url": "http://mwdb.cert.pl/sample/{}".format(job_id),
```
msm-code, line 51: use f-strings

```suggestion
            "url": f"http://mwdb.cert.pl/sample/{job_id}",
```

src/metadata/MwdbIntegrate.py
msm-code: (in config, get api_url from MWDB_API_URL env variable, but use "https://mwdb.cert.pl/api/" as default value)

src/daemon.py
```diff
@@ -97,13 +97,8 @@ def job_daemon() -> None:
                 )
 
         elif queue == "queue-metadata":
-            # LEGACY QUEUE
```
msm-code, line 100: this reverts changes from other PR. You probably resolved merge conflicts by using "keep my changes only" instead of merging properly

src/metadata/MwdbIntegrate.py
```diff
@@ -0,0 +1,55 @@
...
+
+        except:
+            logging.exception("Failed to post mquery metadata")
+            return {}
+
+        obj = {}
+
+        obj["mwdb_sample"] = {
+            "display_text": "mwdb",
+            "url": f"http://mwdb.cert.pl/sample/{job_id}",
```
ITAYC0HEN, line 52: Hardcoded URL for cert.pl's instance. Can it be configured for future open-sourcing of mwdb?

msm-code: I think the whole plugin system needs a revamp, though I'm not sure how to do it in an extensible way. But you're right, I missed that (api url above is configurable but the hyperlink is not).

psrok1:
This PR has many issues:
- file names are not following the convention
- MwdbGetMetadata.py is not using mwdblib
- I don't see mwdblib in requirements, although it's used by MwdbIntegrate (maybe "plugins" like that should have separate requirements file?)
- Hardcoded "mwdb.cert.pl" URL (should be derived from MWDB_API_URL)

I will try to redesign it in #67.

PR #59, opened 2020-04-08
Use ursadb's iterators in mquery

msm-code:
Additionaly, fix things that broke the CI again.
TODO: move the CI pipeline to github...

Reviewable, but should not be merged until ursadb's iterators go through the review.

### DIFF
--- a/src/daemon.py
+++ b/src/daemon.py
@@ -2,23 +2,21 @@
 import json
 import logging
 import time
-
 import yara  # type: ignore
 from functools import lru_cache
-
+import random
 from yara import SyntaxError
-
 import config
 from lib.ursadb import UrsaDb
 from lib.yaraparse import parse_yara, combine_rules
 from util import make_redis, setup_logging
-from typing import Any, Dict, List
+from typing import Any, Dict, List, Optional, Tuple
 
 redis = make_redis()
 db = UrsaDb(config.BACKEND)
 
 
-@lru_cache(maxsize=8)
+@lru_cache(maxsize=32)
 def compile_yara(job_hash: str) -> Any:
     yara_rule = redis.hget("job:" + job_hash, "raw_yara")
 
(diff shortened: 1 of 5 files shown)


src/daemon.py
```diff
@@ -57,62 +55,101 @@ def collect_expired_jobs() -> None:
...
-        queue, data = redis.blpop(
-            ["queue-search", "queue-index", "queue-metadata", "queue-commands"]
-            + yara_queues
-        )
-
-        if queue == "queue-search":
-            job_hash = data
-            logging.info("New task: {}:{}".format(queue, job_hash))
+        if try_to_do_task(not did_anything):
+            logging.info("Task done successfully...")
```
ITAYC0HEN, line 139: Story of my life...

src/lib/ursadb.py
```diff
@@ -25,9 +25,11 @@ def query(self, query: str, taint: str) -> Json:
         start = time.clock()
         if taint:
             taint = taint.replace('"', '\\"')
-            query = 'select with taints ["{}"] {};'.format(taint, query)
+            query = 'select with taints ["{}"] into iterator {};'.format(
```
nazywam, line 28: Any reason for using formats instead of fstrings like in rest of the code?

src/daemon.py
```diff
@@ -148,107 +170,77 @@ def update_metadata(job_hash: str, file_path: str, matches: List[str]) -> None:
...
+    rule = compile_yara(job_hash)
 
-        pipe = redis.pipeline()
-        pipe.rpush("false_positives:" + job_hash, file)
-        pipe.hget("job:{}".format(job_hash), "total_files")
-        pipe.hincrby("job:{}".format(job_hash), "files_processed")
-        _, total_files, files_processed = pipe.execute()
+    for sample in files:
+        try:
+            matches = rule.match(data=open(sample, "rb").read())
```
nazywam, line 194: I guess the FD gets closed after the yara matching right?

btw: the `match` function accepts filepaths as well so no need to read the data ourselves right?

nazywam: ref: https://yara.readthedocs.io/en/v3.4.0/yarapython.html#yara.Rules

src/lib/ursadb.py
msm-code: No, it was just like this before my changes. Fixed!

src/daemon.py
msm-code: Thanks! I didn't notice this, and I think this is unnecessarily slowing down scanning.

Issue #29, opened 2020-04-03
Document the existence of API and its usage


ITAYC0HEN:
**Description**
While it is claimed by @msm-code that mquery has an API, I couldn't find any mention on this in the README nor in the repo. The API should be documented and usage examples should be provided.

msm-code:
We probably should also improve and update ursadb API documentation (though it's usually used together with mquery)

msm-code:
https://cert-polska.github.io/ursadb/docs/
https://cert-polska.github.io/mquery/docs/
https://cert-polska.github.io/mquery/docs/api.html

PR #63, opened 2020-04-08
Migrate to fastapi

BonusPlay:


.github/workflows/test_code.yml
```diff
@@ -10,10 +10,12 @@ jobs:
         uses: actions/setup-python@v1
         with:
           python-version: '3.6'
+      - name: install mypy
+        run: pip3 install mypy==0.770
```
BonusPlay, line 14: it may be a good idea to add here `pydantic` and `fastapi` as they provide mypy types in their modules (which will probably come up as a problem later down the development)

.github/workflows/test_code.yml
```diff
@@ -10,10 +10,12 @@ jobs:
...
+      - name: install requirements
+        run: pip3 install -r requirements.txt
```
BonusPlay, line 16: we don't need to install all dependencies (look above)

msm-code:
Ok, that took way more effor than I've thought, but fixes all regressions (I hope) and dev-docker-compose.

This adds a swagger UI, which may not be a proper documentation, but at least documents api existence (#29):

ss

I've also *tried* not to fix the code in this PR and only migrate it, but there were some race contitions in the old react code that that started showing up on faster webserver - for example on the query page.

src/app.py
```diff
@@ -95,90 +103,94 @@ def query() -> Any:
...
-
-    return jsonify({"job": job, "matches": [json.loads(m) for m in meta]})
+    return MatchesSchema(job=job, matches=[json.loads(m) for m in meta])
+
+
+def get_job(job_id: str) -> JobSchema:
+    job = redis.hgetall(job_id)
+    return JobSchema(
+        id=job_id[4:],
+        status=job.get("status", "[error]"),
```
BonusPlay, line 135: 1) It's a bit counter intuitive to get `"[error"]` in your JSON (string that looks like an array, but contains invalid data). Proposal: `"ERROR"`
2) I don't think returning errors with normal data structure and 200 OK is the way to go (just as a general practice).

src/app.py
```diff
@@ -95,90 +103,94 @@ def query() -> Any:
...
+        total_files=job.get("total_files", 0),
+    )
 
 
-@app.route("/api/job/<hash>")
-def job_info(hash: str) -> Response:
-    return jsonify(redis.hgetall("job:" + hash))
+@app.get("/api/job/{job_id}", response_model=JobSchema)
+def job_info(job_id: str) -> JobSchema:
+    return get_job(f"job:{job_id}")
```
BonusPlay, line 148: I guess this is to reduce API changes, but tbh responding with JSON that has only 1 key is pointless (just return value).

src/app.py
```diff
@@ -95,90 +103,94 @@ def query() -> Any:
...
-        [dict({"id": job[4:]}, **redis.hgetall(job)) for job in jobs],
-        key=lambda o: o.get("submitted"),
-        reverse=True,
-    )
-
-    return jsonify({"jobs": jobs})
+@app.get("/api/job", response_model=JobsSchema)
+def job_statuses() -> JobsSchema:
+    jobs = [get_job(j) for j in redis.keys("job:*")]
+    return JobsSchema(jobs=sorted(jobs, key=lambda j: j.submitted))
```
BonusPlay, line 189: As above, return plain list

src/app.py
msm-code: 1. Right

2. I agree, but redis is schemaless, and I've had situations (not yesterday, just in general) where just one invalid record (for example, missing status) caused 500 erorrs on the whole website. Right now we're using redis so we should just embrace the fact that dhe data may or may not be there. As a longer term goal, I hope we migrate to postgres or some other relational database :thinking:.

.github/workflows/test_code.yml
msm-code: Maybe not all, but I don't want to specify all used modules here one by one (otherwise this will have to be updated for every new typing in our dependencies). Is there any downside to installing them all? (other than execution time)

src/app.py
msm-code: Can you explain? I mean, JobSchema is an object with quite a few elements (including ID, status, etc)

.github/workflows/test_code.yml
BonusPlay: No, just execution time.

src/app.py
BonusPlay: :+1:

src/app.py
msm-code: I'm not too familiar with pydantic's schemas, but JobsSchema is a list of complex objects. How should I change this to better fit into fastapi/pydantic?

src/app.py
BonusPlay: nvm I missread the code ._.

src/app.py
BonusPlay: You don't need to return a pydantic schema. You can just return a `List[JobSchema]` and it will work :)

PR #71, opened 2020-04-09
no-matched-samples-status

### LINKED ISSUE #28: Progress bar shows NaN% when no matches found
**Description**
When no matches and no potential files found, the progress bar can show NaN%. This is probably caused by dividing 0 samples by 0 indexed-potentials in the JS.

Also, notice that the progress bar isn't even fully filled, as expected from a no-results-found job

image

### LINKED ISSUE #32: Show visible indication that no matches were found
**Description**
It is not intuitive to understand that no matches were found for a given rule. The interface shows the table's header (why?) even though there are no matches, the progress bar isn't full, etc.

image

Instead, I suggest to show a full progress bar and then a message saying "No matches found for your rule" instead of the empty table

KrzyWie:
Repair NaN value when no sample was matched and implement info about no results.

resolve #28 resolve #32

### DIFF
--- a/src/mqueryfront/src/QueryStatus.js
+++ b/src/mqueryfront/src/QueryStatus.js
@@ -236,7 +236,6 @@ class QueryStatus extends Component {
         if (isNaN(progress)) {
             progress = 100
             processed = "-"
-
         }
 
         const matches = this.state.matches.map((match, index) => (
@@ -287,7 +286,6 @@ class QueryStatus extends Component {
                 </table>
             )
         }
-
 
         return (
             <div className="mquery-scroll-matches">
--- a/src/mqueryfront/src/QueryStatus.js
+++ b/src/mqueryfront/src/QueryStatus.js
@@ -233,9 +233,10 @@ class QueryStatus extends Component {
             </button>
         );
 
-        if (!this.state.job.total_files) {
-            progress = 0;
-            processed = "-";
+        if (isNaN(progress)) {
+            progress = 100
+            processed = "-"
+
         }
 
         const matches = this.state.matches.map((match, index) => (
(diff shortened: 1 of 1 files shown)


ITAYC0HEN:
closes #28 , closes #32

src/mqueryfront/src/QueryStatus.js
```diff
@@ -233,9 +233,9 @@ class QueryStatus extends Component {
             </button>
         );
 
-        if (!this.state.job.total_files) {
-            progress = 0;
-            processed = "-";
+        if (isNaN(progress)) {
+            progress = 100
```
icedevml, line 237: If `this.state.job.total_files == 0` and the `this.state.job.status != "done"` then I think the progres should be 0 instead of 100 (the job was not actually completed succesfully).

src/mqueryfront/src/QueryStatus.js
```diff
@@ -233,9 +233,9 @@ class QueryStatus extends Component {
             </button>
         );
 
-        if (!this.state.job.total_files) {
-            progress = 0;
-            processed = "-";
+        if (isNaN(progress)) {
```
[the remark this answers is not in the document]
KrzyWie: I started with this condition, however, it was not possible to achieve it, after implementing it this way quinces worked. I will try to implement this condition somehow in a different way.

src/mqueryfront/src/QueryStatus.js
KrzyWie: Technically the job was completed but nothing was found, so the progress is 100, the job is over.

msm-code: what @icedevml means is that total_files may be 0 and increase later, when the job is just starting, right?

msm-code:
Can you rebase your changes on the current master? We've configured CI now, it'd be nice to test this PR with it before merging.

src/mqueryfront/src/QueryStatus.js
icedevml: @msm-code yes, `total_files` is known from certain point, I've referenced appropriate code previously

src/mqueryfront/src/QueryStatus.js
```diff
@@ -265,6 +265,25 @@ class QueryStatus extends Component {
         if (this.state.job.status === "expired") {
             return ReturnExpiredJob(this.state.job.error);
         }
+
+        let results;
```
msm-code, line 269: should this be `<div />` by default? I'm not sure how adding `undefined` element will work

PR #77, opened 2020-04-10
Add fixed version of Malwarecage integration plugin

psrok1:


src/metadata/mwdb_uploads.py
```diff
@@ -0,0 +1,61 @@
...
+
+    def __init__(self, mwdb_api_token: str, mwdb_api_url: str, mwdb_url: str) -> None:
+        super().__init__()
+        self.mwdb = Malwarecage(api_url=mwdb_api_url, api_key=mwdb_api_token)
+        self.mwdb_url = mwdb_url
+
+    def extract(self, matched_fname, current_meta):
+        # '/uploads' Malwarecage directory format
+        # /mnt/samples/9/d/c/5/9dc571ae13a62954155999cae9cecc4f0689e2ba9a8940f81d1e564271507a3e
+        m = re.search(r"/[a-f0-9]/[a-f0-9]/[a-f0-9]/[a-f0-9]/([a-f0-9]+)$", matched_fname)
```
ITAYC0HEN: Depends on how strict you want to be, regex backreference can be used here since, as it seems, the four directories inside "samples" are the first four characters of the checksum. Thus, something like this can be strctier:

```suggestion
        # /mnt/samples/9/d/c/5/9dc571ae13a62954155999cae9cecc4f0689e2ba9a8940f81d1e564271507a3e
        m = re.search(r"/([a-f0-9])/([a-f0-9])/([a-f0-9])/([a-f0-9])/\1\2\3\4([a-f0-9]+)$", matched_fname)
```

See this regex101 for example -> https://regex101.com/r/6G9xUC/2

msm-code: :+1: makes sense

msm-code:
Nitpick: `Added fixed version` -> `Add fixed version` (changed in the description)

src/metadata/mwdb_uploads.py
```diff
@@ -0,0 +1,64 @@
+import re
+import urllib.parse
+
+from metadata import Metadata
+from mwdblib import Malwarecage  # type: ignore
+from typing import List, Any
+
+
+class MalwarecageUploadsMetadata(Metadata):
+    """
```
msm-code, line 10: :+1: docstrings in a standard format are a great idea

src/metadata/mwdb_uploads.py
```diff
@@ -0,0 +1,64 @@
+import re
+import urllib.parse
+
+from metadata import Metadata
+from mwdblib import Malwarecage  # type: ignore
```
msm-code, line 5: I *think* you should add
```
[mypy-mwdblib.*]
ignore_missing_imports = True
```
to setup.cfg instead, and remove the `type: ignore` (@BonusPlay removed all type:ignores from the project that way)

src/metadata/mwdb_uploads.py
psrok1: Changed to:
```python
r"/([a-f0-9])/([a-f0-9])/([a-f0-9])/([a-f0-9])/(\1\2\3\4[a-f0-9]+)$",
```
(whole basename is capturing group)

src/metadata/mwdb_uploads.py
psrok1: Done :+1:

src/metadata/mwdb_uploads.py
psrok1: ^^

PR #85, opened 2020-04-11
Fix service startup order in docker-compose ymls

msm-code:
Try to control service startup order. Otherwise they keep logging connection
errors for a first few seconds. This is not usually a problem (services
will connect eventually), but sifting through irrelevant errors when something
doesn't work may be confusing

This solution is not perfect (it doesn't check if service is ready, only
if the container started running), but it's better than nothing.

### DIFF
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -25,6 +25,7 @@ services:
     - ./src:/usr/src/app/src
     depends_on:
       - "redis"
+      - "ursadb"
   dev-daemon:
     build:
       context: .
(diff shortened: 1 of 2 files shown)


docker-compose.dev.yml
```diff
@@ -21,6 +23,8 @@ services:
     volumes:
     - ./samples:/mnt/samples
     - ./src:/usr/src/app/src
+    depends_on:
+      - "redis"
```
nazywam, line 27: Doesn't dev-web also depend on ursadb?

nazywam:
I guess we don't want to use those `hacky wait-for-postgres.sh`-like scripts?

docker-compose.dev.yml
msm-code: Right, for some reason I've assumed that it passes everything to daemon.

PR #88, opened 2020-04-12
Fix iterator lock handling

msm-code:
Repeat after me: don't return ad-hoc tuples with booleans from critical
functions.

### DIFF
--- a/src/lib/ursadb.py
+++ b/src/lib/ursadb.py
@@ -1,13 +1,26 @@
 import json
 import time
 import zmq  # type: ignore
-from typing import Dict, Any, List, Tuple
+from typing import Dict, Any, List
 
 
 Json = Dict[str, Any]
 
 
-class UrsaDb(object):
+class PopResult:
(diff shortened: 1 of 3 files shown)


src/lib/ursadb.py
```diff
@@ -62,15 +75,13 @@ def pop(self, iterator: str, count: int) -> Tuple[bool, List[str]]:
 
         res = json.loads(response)
         if "error" in res:
```
nazywam, line 77: :thinking:

msm-code: We should probably document ursadb response scheme :thinking:

icedevml: this is prety fragile as it relies on the existence of the key rather than value; also with "statically typed" objects this should be either always false or always true

msm-code: See below:

```python
res["error"].get("retry", False)`
```
So we don't rely only on a key existence, we just treat missing key as False

PR #89, opened 2020-04-12
WIP: Collapsible query and minor UI improvements

ITAYC0HEN:

This PR is a work-in-progress. I applied the prettier for formatting and it looks great on VSCODE, but Github shows huge diff in the code. I'll investigate this tomorrow.

**What is the current behavior?**

- The query pane takes a lot of space when showing the results
- Multiple buttons are too big and steal screen space
- Matched rules are not aligned in results view
- "API Docs" button was stuck in the middle of the screen
- The status label was shown in gray
- Scrollbar contained the progress bar, # of matched samples and status

**What is the new behavior?**

_Click on the GIF to see it in full size_
demo_collapse_query

- Implemented a button to collapse and uncollapse the Query view
  - The collapse button is not visible when no query was executed (e.g writing new query)
  - After clicking the "Query" button, the Query view will collapse since it is less likely to be used
  - When entering an executed query from the /recent page, the Query view will be collapsed
- Implemented a generic function to color both the Status label and the progress bar based on the job's status
- The matched rules labels are now pulled to the right and more spacious
- The buttons in the Query view are smaller now
- The scroll bar is now available only in the Matches table. Hence, the status view with the progress bar 
stays at top
- The "API Docs" button was wrongly aligned. Now it is pulled to the right, assuming the author didn't mean to pull it to the left.

**Test plan**
TODO

**Closing issues**

TODO

### DIFF
--- a/src/mqueryfront/src/App.css
+++ b/src/mqueryfront/src/App.css
@@ -4,10 +4,16 @@
 }
 
 .mquery-scroll-matches {
-    height: calc(100vh - 95px);
+    height: calc(100vh - 235px);
     overflow-y: scroll;
 }
 
 .mquery-scroll-matches td {
     word-break: break-all;
+}
+
+.is-collapsed {
+    -webkit-transform: translateX(-100%);
+    transform: translateX(-100%);
+    position: absolute;
 }
--- a/src/mqueryfront/src/QueryField.js
+++ b/src/mqueryfront/src/QueryField.js
@@ -106,7 +106,7 @@ class QueryField extends Component {
                 <div className="btn-group mb-1" role="group">
                     <button
                         type="button"
-                        className="btn btn-success btn-lg"
+                        className="btn btn-success btn-sm"
                         onClick={(event) =>
                             this.handleQuery(event, "query", "medium")
                         }
@@ -153,7 +153,7 @@ class QueryField extends Component {
                     </div>
                     {this.state.isLocked ? (
                         <button
-                            className="btn btn-secondary btn-lg"
+                            className="btn btn-secondary btn-sm"
                             name="clone"
                             type="submit"
                             onClick={this.handleEdit}
@@ -162,7 +162,7 @@ class QueryField extends Component {
                         </button>
                     ) : (
                         <button
-                            className="btn btn-secondary btn-lg"
+                            className="btn btn-secondary btn-sm"
                             name="parse"
                             type="submit"
                             onClick={(event) =>
--- a/src/mqueryfront/src/QueryPage.js
+++ b/src/mqueryfront/src/QueryPage.js
@@ -17,6 +17,7 @@ class QueryPage extends Component {
 
         this.state = {
             mode: "query",
+            collapsed: false,
             qhash: qhash,
             rawYara: "",
             queryPlan: null,
@@ -27,6 +28,7 @@ class QueryPage extends Component {
         this.updateQhash = this.updateQhash.bind(this);
         this.updateQueryError = this.updateQueryError.bind(this);
         this.updateQueryPlan = this.updateQueryPlan.bind(this);
+        this.collapsePane = this.collapsePane.bind(this);
     }
 
     componentDidMount() {
@@ -66,6 +68,7 @@ class QueryPage extends Component {
             this.props.history.push("/");
         } else {
             this.props.history.push("/query/" + newQhash);
+            this.collapsePane();
         }
 
         this.setState({
@@ -76,7 +79,6 @@ class QueryPage extends Component {
             matches: [],
             job: [],
         });
-
         this.loadMatches();
     }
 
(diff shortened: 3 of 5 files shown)


src/mqueryfront/src/QueryField.js
```diff
@@ -52,7 +52,7 @@ class QueryField extends Component {
                     this.props.updateQhash(
                         response.data.query_hash,
                         this.state.rawYara
-                    );
+                    );
```
msm-code, line 55: Nitpick: unnecessary whitespace (this is also the cause of the failed check)

src/mqueryfront/src/QueryPage.js
```diff
@@ -27,6 +28,7 @@ class QueryPage extends Component {
         this.updateQhash = this.updateQhash.bind(this);
         this.updateQueryError = this.updateQueryError.bind(this);
         this.updateQueryPlan = this.updateQueryPlan.bind(this);
+        this.collapsePane = this.collapsePane.bind(this);
```
msm-code, line 31: shouldn't this be called `togglePane`? (since it's both `collapse` and `uncollapse`)

src/mqueryfront/src/App.css
```diff
@@ -4,10 +4,26 @@
...
-    height: calc(100vh - 95px);
+    height: calc(100vh - 235px);
     overflow-y: scroll;
 }
 
 .mquery-scroll-matches td {
     word-break: break-all;
 }
+
+#query-column {
```
msm-code, line 15: Is it possible to achieve this effect with css classes instead of IDs? AFAIK classes are strongly preferred over IDs in reactjs when possible. I *think* in this case simple class will be enough?

src/mqueryfront/src/App.css
```diff
@@ -4,10 +4,26 @@
...
+    transition: all .3s ease-out;
+}
+
+.is-collapsed {
+    -webkit-transform: translateX(-100%);
+    transform: translateX(-100%);
+    position: absolute;
+}
+
+#results-column {
```
msm-code, line 26: I think it's not used anywhere - can be safely removed

msm-code, line 26: I think it's not used anywhere - can be safely removed

src/mqueryfront/src/QueryPage.js
```diff
@@ -162,16 +170,31 @@ class QueryPage extends Component {
...
+                />
+            </div>
         );
         return (
             <div className="container-fluid">
                 <div className="row">
-                    <div className="col-md-6">
+                    <div
+                        id="query-column"
+                        className={
```
msm-code, line 194: See comment above, I think this ID can be replaced by class (after the CSS update:
```html
                    <div
                        className={
                            "query-column " + 
                            (this.state.collapsed ? "is-collapsed" : "col-md-5")
                        }
                    >
```

src/mqueryfront/src/QueryPage.js
```diff
@@ -182,7 +205,13 @@ class QueryPage extends Component {
                             updateQueryError={this.updateQueryError}
                         />
                     </div>
-                    <div className="col-md-6" id="status-col">
+                    <div
+                        id="results-query"
```
msm-code, line 209: Unused ID, can be removed I think

src/mqueryfront/src/QueryPage.js
```diff
@@ -182,7 +205,13 @@ class QueryPage extends Component {
...
+                        className={
+                            this.state.collapsed ? "col-md-12" : "col-md-7"
+                        }
+                        id="status-col"
```
msm-code, line 213: actually both of them can be removed :thinking:. I don't know why we kept this.

src/mqueryfront/src/QueryPage.js
ITAYC0HEN: That's tricky. Usually, such elements are called "collapsible" in the Bootstrap word. And the button is toggling it. But the functions themselves are called `collapse()`. It is up to you, whatever you decide :)

msm-code: Sounds good, I was making sure I understand the code intention

ITAYC0HEN:
Alright, @msm-code I think it is ready now. All checks are passed.
I removed the animation completely because I couldn't make it work smoothly. I am rusty with boostrap 4 and I think that the mixture of bootsrap3 and 4 in the UI made it more complicated than I think.

So as for now, to collapse is immediate and now animation applied. Thus, no need for element's IDs.
On the same topic, I am not sure I strongly agree with you about not using ID for elements. Unfamiliar with ReactJS best-practices but I don't like creating a class for a single component that will stay single. I am aware fo the debate about using or not using IDs in CSS. But well, no animation - no IDs :)

msm-code:
I see. It looked smooth for me, but I'm the last person you should as so I take your expertise.

Do you have examples of mixed bootstrap3/4 elements? My knowledge of bootstrap is limited to copy&pasting random snippets from the internet until they look good, but I'll happily educate myself and maybe refactor this in a (distant) future.

`I am not sure I strongly agree with you about not using ID for elements` I don't have a *strong* opinion on that topic, just something that I've read about and try to follow. It also makes sense to me as a programmer - ids must be globally unique so eveything with an id is not reusable. Anayway, I wouldn't have a problem with IDs if you insisted, but as you've said - no ids no problem ;)

Issue #42, opened 2020-04-04
Display the number of files that are not indexed


msm-code:
Suggested in #19

> It will also be nice-to-have to show the number of files that aren't indexed because ls /mnt/samples | wc -l != num_of_indexex_files.

Needs backend changes, redesign, and pondering if it's even in scope. I can see how it can be useful though (for example, as an alert when autoindex job fails)

msm-code:
Ok, after a month I still don't have a good idea how to do this efficiently and in a way that will work with s3, multiple agents, multiple ursadb instances, etc. Just counting the files is a very expensive task, so we can't just do it every request, or even every hour.

If that's important to anyone, feel free to reopen and we'll reconsider :thinking:

Issue #35, opened 2020-04-03
Decrease font size on the interface


ITAYC0HEN:
**Description**

By default, when showing the job's results mquery uses 16px font as retrieved from the `:root` by specifying `1rem`. For most places in the interface, 16px is too big and causes a simple `./samples/<sha256>` to line-wrap.

I suggest using a 0.75REM (12px) font when possible.

**Before:**

![image](https://user-images.githubusercontent.com/20182642/78386691-b18da780-75e6-11ea-9562-f23549fe41d3.png)

**After:**

![image](https://user-images.githubusercontent.com/20182642/78386737-c2d6b400-75e6-11ea-86cb-82e1345b9b5e.png)

**Before:**

![image](https://user-images.githubusercontent.com/20182642/78386770-d8e47480-75e6-11ea-8b4f-2e5c6c58ede7.png)

**After:**
![image](https://user-images.githubusercontent.com/20182642/78386804-eac61780-75e6-11ea-89a5-fd20c4e70239.png)

msm-code:
While I agree that UI could use a lot of improvements, I'm not sure that decreasing a font size is the solution. My longer term plan was to ask a real frontend developer to review and improve the UI.

I don't want to randomly mess with the UI in the meantime, because the end result will probably still be ugly :P. In other words, I feel like the frontend needs more design as a whole. Putting this on `priority:low` for now.

msm-code:
@ITAYC0HEN both query view and recent jobs view has changed over the last year:

Recent jobs:
![ss](https://user-images.githubusercontent.com/7026881/144145735-3cd5be71-1eb1-4f92-9312-960afa11dabc.png)

Query view:
![ss](https://user-images.githubusercontent.com/7026881/144145772-03160ca7-4d96-4b28-bcfb-d7aa0e5d2e25.png)

Do you think it is OK to close this issue? I think font size is not the problem anymore. I can always open more tickets about more specific issues.

PR #91, opened 2020-04-13
Feature/#19 my first easy task in mquery

### LINKED ISSUE #19: Show number of total indexed files in the Status page
**Description**
The `/status` page is used to provide information about the underlying infrastructure, including versions, topology, and current connections.
If a user wants to know how many files are indexed, there's no clear way to do so (unless ursadb-cli provides it somehow, and still - it is no accessible).

My suggestion is to display the number of indexed files in the dataset(s). It will also be nice-to-have to show the number of files that aren't indexed because `ls /mnt/samples | wc -l != num_of_indexex_files`.

A bonus feature would be to show a total size of the dataset.

JaroPowerH:

**Test plan**

**Closing issues**

fixes #19

### DIFF
--- a/src/mqueryfront/src/DatabaseTopology.js
+++ b/src/mqueryfront/src/DatabaseTopology.js
@@ -31,6 +31,7 @@ class DatasetRow extends Component {
                     })}
                 </td>
                 <td>{filesize(this.props.size, { standard: "iec" })}</td>
+                <td>{this.props.file_count}</td>
             </tr>
         );
     }
@@ -98,6 +99,7 @@ class DatabaseTopology extends Component {
                                 <th>dataset id</th>
                                 <th>index types</th>
                                 <th>size</th>
+                                <th>file count</th>
                             </tr>
                         </thead>
                         <tbody>{datasetRows}</tbody>


msm-code:
Looks good!

ss

Though not sure about adding a new column - maybe display it in the `size` column instead?

I.e. something like `2.89 GiB (50001 files)`.

Thoughts?

JaroPowerH:
For now, displaying file count in size collumn seems to be better solution.
I can make a change.

Another issue (Display the number of files that are not indexed #42
<https://github.com/CERT-Polska/mquery/issues/42>) will need a place to
display number od file not indexed or total number od files.

We can use collumn "file count" for displaying both values.
I.e. something like 50001 of 58889.

What do you think?

Regards,
JaroPowerH

pon., 13.04.2020, 20:12 użytkownik msm-code <notifications@github.com>
napisał:

ITAYC0HEN:
Thank you! 
So I played with it a bit. I can't tell for sure what data the team prioritizes to show the user at first sight, but this can be a clean solution.

Sod, what do you think about something like this?

mquery_show_topology_collapse2

msm-code:
Ok, maybe first we'll merge the solution from @JaroPowerH and work on the more advanced version that you've shown later.

JaroPowerH:
I will be pleased to work on the solution proposed by ITAYC0HEN.

msm-code:
@ITAYC0HEN mentioned that he has some code already, so maybe  we can leave this to him.

Maybe try #94 or #35

PR #95, opened 2020-04-14
UI improvements to status page

ITAYC0HEN:

This pull request continues the great work done by @JaroPowerH and tries to solve two things.
First, the Topology table in the status is showing four columns. I think that there is extra info in there and sometimes even redundancy.
Second, there is a mixture of Capitalisation and lower-case in titles and table headers.

The PR implements an accordion feature in the table, to not show the index types by default. It also sums up sizes where needed. Trying to focus on what matters to the user.

Secondly, the PR capitalizes the Status page wherever it is needed.

mquery_show_topology_collapse3

Don't get scared by the 5 new CSS properties. They are used to design the table to look better. The creation of a new `td` broke the table-stripped so I had to strip it by myself. It's a common practice.
Another common practice that applied is the toggled arrow in the collapsible rows.

**Test plan**
- Fetch the PR locally and start up mquery
- See that no regression was introduced when displaying the Topology table
- Add multiple indexes to a dataset and see that all of them are clearly visible and that their sizes are correct
- Make sure that no typo was introduce when capitalizing the row headers

**Closing issues**

None

### DIFF
--- a/src/mqueryfront/src/DatabaseTopology.js
+++ b/src/mqueryfront/src/DatabaseTopology.js
@@ -7,8 +7,12 @@ import { API_URL } from "./config";
 
 class DatasetRow extends Component {
     render() {
-        return (
-            <tr>
+        return [
+            <tr
+                data-toggle="collapse"
+                data-target={"#collapsed_" + this.props.id}
+                class="accordion-toggle"
+            >
                 <td>
                     <code>{this.props.id}</code>
                     {this.props.taints.map((taint) => (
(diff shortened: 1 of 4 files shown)


msm-code:
Hi,

again, thanks for the contribution!

What do you think about hinting what the indexes are before uncollapsing?

Before clicking:

ss

After clicking:

ss

I'm not saying it's a good idea, just thinking aloud.

Why I think it may be useful? Because some operations are only possible on datasets with the same indexes (for example compacting). Expanding all the datasets to see how compactible database is will take a lot of clicks in case there are a lot of datasets.

ITAYC0HEN:
too noisy in my opinion. If compacting is only applicable for some indexes, than don't show it when it's not applicable.

Also, if there's a way to do calculation of how useful compacting is, then do it for the user :)

src/mqueryfront/src/DatabaseTopology.js
```diff
@@ -7,8 +7,12 @@ import { API_URL } from "./config";
 
 class DatasetRow extends Component {
     render() {
-        return (
-            <tr>
+        return [
+            <tr
+                data-toggle="collapse"
+                data-target={"#collapsed_" + this.props.id}
```
msm-code, line 13: This should **probably** be done with refs (yeah, my crusade against ids again), but after taking a 15s look at it I think I don't think it's worth it (more problems than it's worth right now).

msm-code:
```
too noisy in my opinion. If compacting is only applicable for some indexes, than don't show it when it's not applicable.

Also, if there's a way to do calculation of how useful compacting is, then do it for the user :)
```

Yeah, that's what I think too. I see many potentnial improvements here:
 - for advanced users, option to select multiple datasets with checkboxes on the left and compact them together
 - for normal users, let mquery worry about compacting (without manual guidance most DBs will end up in a sorry state, like your instance :P)
 - display locked datasets (ones that are currecly being compacted)

But all of these are somethign that neds more design and are a material for the future PR probably (maybe I'll create an issue)

Issue #39, opened 2020-04-04
Add pagination or lazy-load to results table


ITAYC0HEN:
**Description**
When executing a query with many results, the interface will show all the matches in the results table. The table can be with tens of thousands of results for a big dataset, which will make the page slow to response. I suggest implementing pagination (the xhr requests are already fetching 50 at a time) or a [lazy-loading](https://en.wikipedia.org/wiki/Lazy_loading)

msm-code:
Yeah, that can be quite a pain

msm-code:
To test:

1. Fenerate a few test files:

```python
import string
import os
import random

for i in range(40000):
    print(i)
    data = os.urandom(40000)
    open(os.urandom(16).hex(), 'wb').write(data)
```

2. index them all

3. Do some trivial query like:

```
rule hmm {
strings: $a = "a"
condition: $a
}
```

4. Observe that the query finishes quickly, but loading all the results takes ages.

ITAYC0HEN:
#78 is somewhat related to the planning of the new table. A modern approach to a filterable table.

msm-code:
fixed with #96

Issue #37, opened 2020-04-04
Mquery is counting results when displaying previous jobs


ITAYC0HEN:
**Description**
When entering previous jobs, mquery is counting the number of matches instead of showing the number immediately.

![mquery keep counting](https://user-images.githubusercontent.com/20182642/78422224-73d06380-7666-11ea-9685-bede94991e4b.gif)

**Expected behavior**
I expect the total number of results to be shown imeediately. Even if the table is needed to be fetched (lazy loading / pagination will be better btw)

msm-code:
Fixed with #96

PR #96, opened 2020-04-15
add pagination to results table

KrzyWie:
Close issue #39 #37

### DIFF
--- a/src/mqueryfront/src/QueryPage.js
+++ b/src/mqueryfront/src/QueryPage.js
@@ -81,6 +81,7 @@ class QueryPage extends Component {
             matches: [],
             job: [],
         });
+        this.loadJob();
         this.loadMatches();
     }
 
@@ -90,10 +91,60 @@ class QueryPage extends Component {
         
     }
 
+
+    loadJob() {
+        const LIMIT = 50;
+
+        if (!this.state.qhash) {
+            return;
+        }
+
+        axios
+            .get(
+                API_URL +
+                    "/matches/" +
+                    this.state.qhash +
+                    "?offset=" +
+                    0 +
+                    "&limit=" +
+                    LIMIT
+            )
+            .then((response) => {
+                let newShouldRequest = true;
+
+                if (
+                    ["done", "cancelled", "failed", "expired"].indexOf(
+                        response.data.job.status
+                    ) !== -1
+                ) {
+                    if (response.data.job.files_processed >= response.data.job.total_files) {
+                        newShouldRequest = false;
+                    }
+                }
+
+                this.setState({
+                    job: response.data.job,
+                });
+
+                if (newShouldRequest) {
+                    let nextTimeout =
+                        response.data.matches.length >= LIMIT ? 50 : 1000;
+                    this.timeout = setTimeout(
+                        () => this.loadJob(),
+                        nextTimeout
+                    );
+                }
+            })
+            .catch(() => {
+                this.setState({
+                    shouldRequest: false,
+                });
+            });
+    }
+
     loadMatches() {
         const LIMIT = 20;
         let OFFSET = (this.state.activePage - 1) * 20 + 1
-
         axios
             .get(
                 API_URL +
@@ -105,37 +156,11 @@ class QueryPage extends Component {
                     LIMIT
             )
             .then((response) => {
-                // let newShouldRequest = true;
-
-                // if (
-                //     ["done", "cancelled", "failed", "expired"].indexOf(
-                //         response.data.job.status
-                //     ) !== -1
-                // ) {
-                //     if (!response.data.matches.length) {
-                //         newShouldRequest = false;
-                //     }
-                // }
-
                 this.setState({
                     matches: response.data.matches,
-                    job: response.data.job,
                 });
 
-                // if (newShouldRequest) {
-                //     let nextTimeout =
-                //         response.data.matches.length >= LIMIT ? 50 : 1000;
-                //     this.timeout = setTimeout(
-                //         () => this.loadMatches(),
-                //         nextTimeout
-                //     );
-                // }
-            })
-            // .catch(() => {
-            //     this.setState({
-            //         shouldRequest: false,
-            //     });
-            // });
+             })
     }
 
     updateQueryError(newError, rawYara) {
--- a/src/mqueryfront/src/QueryPage.js
+++ b/src/mqueryfront/src/QueryPage.js
@@ -97,45 +97,24 @@ class QueryPage extends Component {
                     "/matches/" +
                     this.state.qhash +
                     "?offset=" +
-                    1 +
+                    0 +
                     "&limit=" +
                     LIMIT
             )
             .then((response) => {
-                let newShouldRequest = true;
-
-                if (
-                    ["done", "cancelled", "failed", "expired"].indexOf(
-                        response.data.job.status
-                    ) !== -1
-                ) {
-                    if (
-                        response.data.job.files_processed >=
-                        response.data.job.total_files
-                    ) {
-                        newShouldRequest = false;
-                    }
-                }
-
+                let job = response.data.job;
                 this.setState({
-                    job: response.data.job,
+                    job: job,
                     matches: response.data.matches,
                 });
-
-                if (newShouldRequest) {
-                    let nextTimeout =
-                        response.data.job.files_matched >= LIMIT ? 20 : 1000;
-                    this.timeout = setTimeout(
-                        () => this.loadJob(),
-                        nextTimeout
-                    );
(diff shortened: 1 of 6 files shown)


ITAYC0HEN:
For some reason, I can't make it work with `docker-compose.dev.yml`:
```
dev-frontend_1  | Starting the development server...
dev-frontend_1  | 
dev-frontend_1  | Failed to compile.
dev-frontend_1  | 
dev-frontend_1  | ./src/QueryResultsStatus.js
dev-frontend_1  | Module not found: Can't resolve 'react-js-pagination' in '/app/src'

```

but it does work with the regular one :) The solution looks good in terms of UI, and anything else :D\
I want to test it when running a query but all my 3 instances are broken now :(


Good work!

msm-code:
@ITAYC0HEN had the same problem, you need to do a rebuild first (`docker-compose -f docker-compose.dev.yml build`). That's because this PR adds a new npm dependency (react-js-pagination).

msm-code:
Looks good!

But there are a few things that don't make sense after this change:

- query page will still try to load every result in memory (I've opened the page, and it downloaded all 10k results even though I just stayed on the first page) - it should be downloaded lazily on-demand

- query page should display the number of results instead of displaying numbre of downloaded results. This is already a part of a different issue ( #37 ), so if you don't feel like it feel free to fix it in a different PR.

ITAYC0HEN:
> @ITAYC0HEN had the same problem, you need to do a rebuild first (`docker-compose -f docker-compose.dev.yml build`). That's because this PR adds a new npm dependency (react-js-pagination).

Thanks, fixed this :)

ITAYC0HEN:
If you're up for lazy load then you can reconsider pagination in favor of. lazy load scrolling (like Facebook, Twitter feeds, VT results)

msm-code:
You forgot mwdb!

Yeah, that's a viable solution too. Anything that will make it possible to reduce a number of queries to the backend for a finished large query.

I prefer paging, but I can see lazy load working here too.

src/mqueryfront/src/QueryPage.js
```diff
@@ -90,10 +91,60 @@ class QueryPage extends Component {
...
+                    this.state.qhash +
+                    "?offset=" +
+                    0 +
+                    "&limit=" +
+                    LIMIT
+            )
+            .then((response) => {
+                let newShouldRequest = true;
+
+                if (
```
msm-code: maybe move setState above this, and return in if, instead of using a temprary vraiable here?

src/mqueryfront/src/QueryPage.js
```diff
@@ -90,10 +91,60 @@ class QueryPage extends Component {
...
+                });
+
+                if (newShouldRequest) {
+                    let nextTimeout =
+                        response.data.matches.length >= LIMIT ? 50 : 1000;
+                    this.timeout = setTimeout(
+                        () => this.loadJob(),
+                        nextTimeout
+                    );
+                }
```
msm-code: All my suggestions in one comment:

```suggestion
                let job = response.data.job;
                this.setState({
                    job: job,
                    matches: response.data.matches,
                });

                let doneStatuses = ["done", "cancelled", "failed", "expired"];
                let isDone = doneStatuses.indexOf(job.status) !== -1;
                let processedAll = job.files_processed >= job.total_files;
                if (isDone && processedAll) {
                    return;
                }

                this.timeout = setTimeout(() => this.loadJob(), 1000);
```

msm-code: After this rewrite it also shows a bug (that I think was there even before your changes) - if the job is cancelled/failed but it didn't process all files, mquery will keep querying the backend. Maybe it should be `||` instead of `&&`?

src/mqueryfront/src/QueryPage.js
```diff
@@ -23,6 +23,8 @@ class QueryPage extends Component {
             queryPlan: null,
             queryError: null,
             datasets: {},
+            activePage: 1,
+            newShouldRequest: true,
```
msm-code, line 27: Is this variable used by anything?

src/mqueryfront/src/QueryPage.js
```diff
@@ -95,7 +97,7 @@ class QueryPage extends Component {
                     "/matches/" +
                     this.state.qhash +
                     "?offset=" +
-                    this.state.matches.length +
+                    1 +
```
msm-code, line 100: Offset should be 0

src/mqueryfront/src/QueryResultsStatus.js
```diff
@@ -63,13 +64,29 @@ class QueryResultsStatus extends Component {
     constructor(props) {
         super(props);
 
+        this.state = {
+            activePage: 1,
+            itemPerPage: 20,
```
msm-code, line 69: Plural:
```suggestion
            itemsPerPage: 20,
```

src/mqueryfront/src/QueryResultsStatus.js
```diff
@@ -142,12 +159,21 @@ class QueryResultsStatus extends Component {
...
         }
 
-        const lenMatches = this.props.matches.length;
+        const lenMatches = this.props.job.files_matched;
 
         if (this.props.job.status === "expired") {
             return ReturnExpiredJob(this.props.job.error);
         }
 
+        // var indexOfLastMatch = this.state.activePage * this.state.itemPerPage;
```
msm-code, line 168: Remove unnecessary comments

src/mqueryfront/src/QueryResultsStatus.js
```diff
@@ -164,6 +190,17 @@ class QueryResultsStatus extends Component {
                         </thead>
                         <tbody>{matches}</tbody>
                     </table>
+                    <div>
+                        <Pagination
+                            activePage={this.state.activePage}
+                            itemsCountPerPage={this.state.itemPerPage}
```
msm-code, line 196:
```suggestion
                            itemsCountPerPage={this.state.itemsPerPage}
```

src/mqueryfront/src/QueryResultsStatus.js
```diff
@@ -63,13 +64,29 @@ class QueryResultsStatus extends Component {
...
     }
 
+    sendResultsActivePage = (pageNumber) => {
+        this.props.parentCallback(pageNumber);
+    }
+
+    handlePageChange(pageNumber) {
+        this.setState({ activePage: pageNumber });
+        this.sendResultsActivePage(pageNumber);
+
```
msm-code, line 87: unnecessary empty line

src/mqueryfront/src/QueryPage.js
```diff
@@ -133,6 +138,32 @@ class QueryPage extends Component {
...
 
+    callbackResultsActivePage = (pageNumber) => {
+        this.setState({ activePage: pageNumber }, () => {
+            this.loadMatches();
+        });
+    };
+
+    loadMatches() {
+        const LIMIT = 20;
+        let OFFSET = (this.state.activePage - 1) * 20 + 1;
```
msm-code, line 149: Why `+1`?

msm-code:
Nitpick: doing a new query should reset active page number (if i go to the 8th page, run a new query, pagination will automatically start at the 8th page)

msm-code:
And also: please rebase your changes on the master, since there are merge conflicts

src/mqueryfront/src/QueryResultsStatus.js
```diff
@@ -164,6 +184,19 @@ class QueryResultsStatus extends Component {
                         </thead>
                         <tbody>{matches}</tbody>
                     </table>
+                    {lenMatches > 20 && (
+                        <div>
```
msm-code, line 188: I think div is unnecessary here?

src/mqueryfront/src/QueryPage.js
```diff
@@ -95,40 +98,48 @@ class QueryPage extends Component {
...
-                    let nextTimeout =
-                        response.data.matches.length >= LIMIT ? 50 : 1000;
-                    this.timeout = setTimeout(
-                        () => this.loadMatches(),
-                        nextTimeout
-                    );
+                let doneStatuses = ["done", "cancelled", "failed", "expired"];
+                let isDone = doneStatuses.indexOf(job.status) !== -1;
+                let processedAll = job.files_processed >= job.total_files;
+                if (isDone && processedAll) {
```
msm-code, line 114: Still `&&`? If nothing else changed, this will cause infinite loop of requests.

msm-code: I've also confirmed it locally. Test plan:

- start a big query
- stop it in a middle such that files_processed < total_files (remember that they're strings)
- observe that requests are repeated in the loop every 1s

PR #98, opened 2020-04-15
Add string xor modifier support

### LINKED ISSUE #51: Research the usefulness of accelerating xor modifier
Accelerate rules like:

```
rule XorExample1
{
    strings:
        $xor_string = "This program cannot" xor

    condition:
       $xor_string
}
```

My gut feeling is that this may result in way too many false positives, but I may be wrong.

We should probably transform strings like this to an OR of 256 xored strings in the backend. I don't feel like first-class support in ursadb is necessary (unless someone finds a way to optimise this that needs it).

nazywam:

Yara strings with a `xor` modifiers are ignored

The yara->ursa parser replaces the xor-modified string with an `or` expression containing all possible cases

**Test plan**

**Closing issues**

fixes #51

### DIFF
--- a/src/lib/yaraparse.py
+++ b/src/lib/yaraparse.py
@@ -158,6 +158,7 @@ def ursify_xor_string(string: PlainString) -> UrsaExpression:
     text_ascii = string.pure_text
     xored_strings: List[UrsaExpression] = []
 
+    # TODO implement modifier ranges - https://github.com/CERT-Polska/mquery/issues/100
     for xor_key in range(256):
         xored_ascii = xor(text_ascii, bytes([xor_key]))
         xored_wide = bytes(x ^ xor_key for y in text_ascii for x in [y, 0])
--- a/src/lib/yaraparse.py
+++ b/src/lib/yaraparse.py
@@ -160,7 +160,7 @@ def ursify_xor_string(string: PlainString) -> UrsaExpression:
 
     for xor_key in range(256):
         xored_ascii = xor(text_ascii, bytes([xor_key]))
-        xored_wide = bytes(x for y in xored_ascii for x in [y, 0])
+        xored_wide = bytes(x ^ xor_key for y in text_ascii for x in [y, 0])
 
         if string.is_ascii:
             xored_strings.append(UrsaExpression(f"{{{xored_ascii.hex()}}}"))
--- a/src/lib/yaraparse.py
+++ b/src/lib/yaraparse.py
@@ -1,26 +1,32 @@
 import argparse
-from yaramod import (  # type: ignore
-    Yaramod,
+import itertools
+import re
+from typing import Dict, List, Optional
+
+from yaramod import (
     AndExpression,
-    StringCountExpression,
-    IntLiteralExpression,
-    ParenthesesExpression,
-    GtExpression,
     EqExpression,
-    OrExpression,
-    StringExpression,
-    OfExpression,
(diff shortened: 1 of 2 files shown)


nazywam:
I've ran into some issues with the `xor` modifiers.
It turns out that yara allows a sub-modifiers for `xor` like this:

* `"hello world" xor(0x1f-0x20)`
* `"hello world" xor(34)`

They are parsed correctly by yaramod (version upgrade was needed) but as far as I see there is no way to read the using the python bindings: [code](https://github.com/avast/yaramod/blob/master/src/python/yaramod_python.cpp#L226)

One way of resolving this issue if parsing the `modifiers_text` ourselves but given the possibility of providing the xor range using both decimal and hex I'd really want to avoid this.

@msm-code can you weigh in on this?

msm-code:
I think you're right that there is currently no way to get the modifiers out of the python bindings.

I think right now we can ignore the modifiers, and maybe we can submit a PR (or issue) upstream to fix this. After a few weeks it should land in pypi and we'll use this feature in mquery too.

Thoughts?

nazywam:
>  we can submit a PR (or issue) upstream to fix

+1

Always wanted to learn how those binding thingies worked

src/lib/yaraparse.py
```diff
@@ -148,11 +154,30 @@ def ursify_plain_string(string: PlainString) -> UrsaExpression:
...
+def ursify_xor_string(string: PlainString) -> UrsaExpression:
+    text_ascii = string.pure_text
+    xored_strings: List[UrsaExpression] = []
+
+    for xor_key in range(256):
+        xored_ascii = xor(text_ascii, bytes([xor_key]))
+        xored_wide = bytes(x ^ xor_key for y in text_ascii for x in [y, 0])
+
+        if string.is_ascii:
+            xored_strings.append(UrsaExpression(f"{{{xored_ascii.hex()}}}"))
```
msm-code, line 166: created an issue to refactor it later #99 (also it's a good candidate for a first issue for someone)

src/lib/yaraparse.py
```diff
@@ -148,11 +154,30 @@ def ursify_plain_string(string: PlainString) -> UrsaExpression:
         return ursa_ascii
 
 
+def ursify_xor_string(string: PlainString) -> UrsaExpression:
+    text_ascii = string.pure_text
+    xored_strings: List[UrsaExpression] = []
+
+    for xor_key in range(256):
```
msm-code, line 161: Maybe add a comment like

```
# TODO(#100) implement modifier ranges

File: src/lib/yaraparse.py [hunk 3]
```diff
@@ -148,11 +154,30 @@ def ursify_plain_string(string: PlainString) -> UrsaExpression:
...
+        if string.is_wide:
+            xored_strings.append(UrsaExpression(f"{{{xored_wide.hex()}}}"))
+
+    return UrsaExpression.or_(*xored_strings)
+
+
 def ursify_string(string: String) -> Optional[UrsaExpression]:
-    if string.is_xor or string.is_nocase:
+    # nocase strings are not currently supported
+    if string.is_nocase:
```
msm-code, line 175: at_least_not_yet

src/lib/yaraparse.py
msm-code: (#100 is the issue number. Not sure about the reference, we can also use a full link to be more explicit)

Issue #74, opened 2020-04-10
Migrate the backend database to postgres


msm-code:
We're using redis for historic reasons, but we should probably move the metadata to postgres, to ensure data integrity and maybe even improve performance.

- [ ] Design a schema and update docker-compose files DB;
- [ ] Change all uses of redis in the code to postgres;
- [ ] Provide a migration scripts?
- [ ] Update documentation (including installation methods)

This is pretty low priority because it's a lot of work and don't add any immediate features though

Suggested by @BonusPlay (not directly, but related to schema problems)

ITAYC0HEN:
On one hand you say,
> and don't add any immediate features

OTOH you also mentioned
> ...and maybe even improve performance.

Can you measure or guess what is the expected performence improvements, if any?
If it is meaningful, the priority rises, isn't it? It if it not meaningful, then yes... sure

msm-code:
Yeah, we could measure performance after the changes. OTOH i expect improvements, if any, to be be pretty small. I think the huge wins from this change is making the system a bit more stable and easier to understand (for example, right now it's not obvious what exactly is stored in `job:xxx` or `meta:xxx` objects).

msm-code:
And there's one more thing. We had to implement a GC for query results, because we need to store matched filenames somewhere (in the db = redis in our case), and redis wants to to fit all the data into memory. After some time this may exhaust the server memory, so right now we remove stale results after some time. With postgres, disk space is a limit.

msm-code:
I still like Postgres, but can't justify working on this issue:
 - I don't know about any users that have performance problems because of this (if you're one, please let me know). There are bottleneck, but redis is not one.
 - This will make upgrade **way** harder for current users.
 - This is a lot of work, and my spare time is (unfortunately) finite. So I'd like to focus on issues with a better effort/benefit ratio.

msm-code:
The current plan:

* [ ] Create a MVP that will use a database for something simple (like configuration) and pass the e2e tests. Update the installation documentation. #367 
* [ ] Migrate all the persistent objects to this layer, while keeping the db.Database shim (don't make changes in random places in code yet)
* [ ] Refactor the inenvitable behemoth - with ORM we may consider moving away from the one central database interface.

PR #104, opened 2020-04-17
Refactor redis everywhere into one db.py class

msm-code:

Right now the data handling is a total fragile mess. If we want
to add any more features we need to fix it. As a first step, I've moved
all operations touching redis to a separate file (db.py)

Redis is safely contained in db.py.

**Test plan**

Just use mquery like normally, and ensure that nothing crashes.

**Closing issues**

A step towards #74

### DIFF
--- a/src/daemon.py
+++ b/src/daemon.py
@@ -34,14 +34,10 @@ def collect_expired_jobs() -> None:
         return
 
     exp_time = int(60 * config.JOB_EXPIRATION_MINUTES)  # conversion to seconds
-    job_ids: List[JobId] = []
 
     for job in redis.get_job_ids():
-        job_ids.append(job)
-
-    for job in job_ids:
-        job_submitted_time = redis.get_job_submitted(job)
-        if (int(time.time()) - job_submitted_time) >= exp_time:
+        job_submission_time = redis.get_job_submitted(job)
+        if (int(time.time()) - job_submission_time) >= exp_time:
             redis.expire_job(job)
 
 
@@ -62,10 +58,10 @@ def process_task(queue: str, data: str) -> None:
 
 
 def try_to_do_task() -> bool:
-    queue, task = redis.get_task()
-    if task is not None:
-        data = task
-        process_task(str(queue), str(data))
+    queue_and_task = redis.get_task()
+    if queue_and_task is not None:
+        queue, task = queue_and_task
+        process_task(queue, task)
         return True
 
     return False
--- a/src/daemon.py
+++ b/src/daemon.py
@@ -11,13 +11,13 @@ from util import setup_logging
 from typing import Any, Dict, List
 from db import Database, JobId, MatchInfo
 
-redis = Database()
-db = UrsaDb(config.BACKEND)
+db = Database()
+ursa = UrsaDb(config.BACKEND)
 
 
 @lru_cache(maxsize=32)
 def compile_yara(job: JobId) -> Any:
-    yara_rule = redis.get_yara_by_job(job)
+    yara_rule = db.get_yara_by_job(job)
 
     logging.info("Compiling Yara")
     try:
@@ -35,10 +35,10 @@ def collect_expired_jobs() -> None:
 
     exp_time = int(60 * config.JOB_EXPIRATION_MINUTES)  # conversion to seconds
 
-    for job in redis.get_job_ids():
-        job_submission_time = redis.get_job_submitted(job)
+    for job in db.get_job_ids():
+        job_submission_time = db.get_job_submitted(job)
         if (int(time.time()) - job_submission_time) >= exp_time:
-            redis.expire_job(job)
+            db.expire_job(job)
 
 
 def process_task(queue: str, data: str) -> None:
@@ -50,15 +50,15 @@ def process_task(queue: str, data: str) -> None:
             execute_search(job)
         except Exception as e:
             logging.exception("Failed to execute job.")
-            redis.fail_job(None, job, str(e))
+            db.fail_job(None, job, str(e))
     elif queue == "queue-commands":
         logging.info("Running a command: %s", data)
-        resp = db.execute_command(data)
+        resp = ursa.execute_command(data)
         logging.info(resp)
 
 
 def try_to_do_task() -> bool:
-    queue_and_task = redis.get_task()
+    queue_and_task = db.get_task()
     if queue_and_task is not None:
         queue, task = queue_and_task
         process_task(queue, task)
@@ -68,17 +68,17 @@ def try_to_do_task() -> bool:
 
 
 def try_to_do_search() -> bool:
-    rnd_job = redis.get_random_job_by_priority()
+    rnd_job = db.get_random_job_by_priority()
     if rnd_job is None:
         return False
     yara_list, job = rnd_job
-    job_data = redis.get_job(job)
+    job_data = db.get_job(job)
 
     try:
         BATCH_SIZE = 500
         if job_data.iterator is None:
             raise RuntimeError(f"Job {job} has no iterator")
-        pop_result = db.pop(job_data.iterator, BATCH_SIZE)
+        pop_result = ursa.pop(job_data.iterator, BATCH_SIZE)
         if pop_result.was_locked:
             return True
         if pop_result.files:
@@ -89,10 +89,10 @@ def try_to_do_search() -> bool:
                 job_data.iterator,
                 job,
             )
-            redis.finish_job(yara_list, job)
+            db.finish_job(yara_list, job)
     except Exception as e:
         logging.exception("Failed to execute yara match.")
-        redis.fail_job(yara_list, job, str(e))
+        db.fail_job(yara_list, job, str(e))
     return True
 
 
@@ -102,7 +102,7 @@ def job_daemon() -> None:
 
     for extractor in config.METADATA_EXTRACTORS:
         logging.info("Plugin loaded: %s", extractor.__class__.__name__)
-        extractor.set_redis(redis.unsafe_get_redis())
+        extractor.set_redis(db.unsafe_get_redis())
 
     logging.info("Daemon loaded, entering the main loop...")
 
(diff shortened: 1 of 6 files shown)


src/db.py
```diff
@@ -0,0 +1,250 @@
...
+        p = self.redis.pipeline(transaction=False)
+        p.hgetall(job.key)
+        p.lrange("meta:" + job.hash, offset, offset + limit - 1)
+        job, meta = p.execute()
+        return MatchesSchema(job=job, matches=[json.loads(m) for m in meta])
+
+    def run_command(self, command: str) -> None:
+        self.redis.rpush("queue-commands", command)
+
+    def get_task(self) -> Tuple[Optional[str], Optional[str]]:
```
BonusPlay, line 223: Wouldn't be better to return `Optional[Tuple[str, str]]`?

src/db.py
```diff
@@ -0,0 +1,250 @@
...
+
+class MatchInfo:
+    """ Represents information about a single match """
+
+    def __init__(self, file: str, meta: Dict[str, Any], matches: List[str]):
+        self.file = file
+        self.meta = meta
+        self.matches = matches
+
+    def to_json(self):
```
BonusPlay, line 69: `-> str` ?

src/db.py
```diff
@@ -0,0 +1,250 @@
...
+        self.redis.hset(job.key, "status", "expired")
+        self.redis.delete(job.meta_key)
+
+    def fail_job(self, queue: Optional[JobQueue], job: JobId, message: str):
+        """ Sets the job status to failed, and removes it from job queues """
+        self.redis.hmset(job.key, {"status": "failed", "error": message})
+        if queue:
+            self.redis.lrem(queue.name, 0, job.hash)
+
+    def cancel_job(self, job: JobId):
```
BonusPlay, line 107: `-> None` ?

src/db.py
```diff
@@ -0,0 +1,250 @@
...
+    def get_job_ids(self) -> List[JobId]:
+        """ Gets IDs of all jobs in the database """
+        return [JobId(key) for key in self.redis.keys("job:*")]
+
+    def expire_job(self, job: JobId) -> None:
+        """ Sets the job status to expired, and removes it from the db """
+        self.redis.hset(job.key, "status", "expired")
+        self.redis.delete(job.meta_key)
+
+    def fail_job(self, queue: Optional[JobQueue], job: JobId, message: str):
```
BonusPlay, line 101: `-> None` ?

src/db.py
```diff
@@ -0,0 +1,250 @@
...
+
+    def __init__(self, key: str) -> None:
+        """ Creates a new JobId object. Can take both key and raw hash. """
+        if not key.startswith("job:"):
+            key = f"job:{key}"
+        self.key = key
+        self.hash = key[4:]
+
+    @property
+    def meta_key(self):
```
BonusPlay, line 38: `-> str` ?

src/db.py
```diff
@@ -0,0 +1,250 @@
...
+
+    def to_json(self):
+        """ Converts match info to json """
+        return json.dumps(
+            {"file": self.file, "meta": self.meta, "matches": self.matches}
+        )
+
+
+class Database:
+    def __init__(self):
```
nazywam, line 77:
```suggestion
    def __init__(self) -> None:
```

src/daemon.py
```diff
@@ -30,100 +29,74 @@ def compile_yara(job_hash: str) -> Any:
...
     if config.JOB_EXPIRATION_MINUTES <= 0:
         return
 
     exp_time = int(60 * config.JOB_EXPIRATION_MINUTES)  # conversion to seconds
-    job_hashes = []
+    job_ids: List[JobId] = []
 
-    for job_hash in redis.keys("job:*"):
-        job_hashes.append(job_hash[4:])
+    for job in redis.get_job_ids():
```
nazywam, line 39: Is `job_ids` used anywhere else besides the loop below? Seems like it isn't really needed

src/daemon.py
```diff
@@ -30,100 +29,74 @@ def compile_yara(job_hash: str) -> Any:
...
+        job_ids.append(job)
 
-    for job in job_hashes:
-        redis.set("gc-lock", "locked", ex=60)
-        job_submitted_time = int(redis.hget("job:" + job, "submitted"))
+    for job in job_ids:
+        job_submitted_time = redis.get_job_submitted(job)
```
nazywam, line 43: Nitpicking but perhaps `job_submission_time`?

src/daemon.py
```diff
@@ -30,100 +29,74 @@ def compile_yara(job_hash: str) -> Any:
...
-    for queue in task_queues:
-        task = redis.lpop(queue)
-        if task is not None:
-            data = task
-            process_task(queue, data)
-            return True
+    queue, task = redis.get_task()
+    if task is not None:
+        data = task
+        process_task(str(queue), str(data))
```
nazywam, line 68: Just making sure, are the `str`s are here on purpose?
`get_task()` might return None values so process_task might be ran like `process_task("None", something)`

src/daemon.py
```diff
@@ -1,24 +1,23 @@
...
 from lib.ursadb import UrsaDb
 from lib.yaraparse import parse_yara, combine_rules
-from util import make_redis, setup_logging
-from typing import Any, Dict, List, Optional, Tuple
+from util import setup_logging
+from typing import Any, Dict, List
+from db import Database, JobId, MatchInfo
 
-redis = make_redis()
+redis = Database()
```
nazywam, line 14: nitpicking again, but won't the `redis` variable name get confused with the redis connection?

src/db.py
nazywam, line 38:
```suggestion
    def meta_key(self) -> str:
```

src/db.py
nazywam, line 69:
```suggestion
    def to_json(self) -> str:
```

src/db.py
nazywam, line 101:
```suggestion
    def fail_job(self, queue: Optional[JobQueue], job: JobId, message: str) -> None:
```

src/db.py
nazywam, line 107:
```suggestion
    def cancel_job(self, job: JobId) -> None:
```

src/db.py
```diff
@@ -0,0 +1,250 @@
...
+
+    def get_task(self) -> Tuple[Optional[str], Optional[str]]:
+        task_queues = ["queue-search", "queue-commands"]
+        for queue in task_queues:
+            task = self.redis.lpop(queue)
+            if task is not None:
+                return queue, task
+        return None, None
+
+    def unsafe_get_redis(self) -> Any:
```
nazywam, line 231:
```suggestion
    def unsafe_get_redis(self) -> StrictRedis:
```

nazywam:
Not exactly sure if this is this pull requests fault but I've found that I'm unable to cancel jobs that are very early in the processing stage.

Reproduction steps:

1. Launch the query with any YARA rule
2. Click the `cancel` button as soon as possible
3. The job status changes to `cancelled` 
image

4. But if we look in recent jobs tab we'll find that it's still in progress/done:
image

src/daemon.py
msm-code: Actually, the code is not supposed to use redis connection directly from this class. So it shouldn't get confused :thinking: Does it make sense?

src/daemon.py
msm-code: Do you mean, why is it a temporary varluable variable instead of `for job in redis.get_job_ids()`? I assumed it's to avoid race conditions of some kind, but on a second thought it doesn't make sense. I tried to avoid doing any logical changes, but in this case it may make sense :thinking: .

src/daemon.py
msm-code: English is hard

src/daemon.py
msm-code: Actually it can't, because it will either return `None, None` or `str, str`. I've refactored this like it was suggested elsewhere with the other suggestion, so they are not necessary anymore.

src/db.py
msm-code: It would, I just tried to avoid doing oto many changes (and also was lazy). Fixed now.

msm-code:
@nazywam I don't think it's introduced in this PR (I've tried not to change any logic), but the job queue in the backend has a pretty low quality (even with a random sleep in a loop sonewhere) so it's probably to blame.

I'll take a look at refactoring it later, for now I've created an issue #108

src/daemon.py
nazywam: I'm also speaking from my own experience, at the first read-through of the changes I was surprised that the redis connection has so many interesting methods.

src/daemon.py
nazywam: Oh so the tostring methods were added just to stop the linter from complaining? 😈

> Actually it can't, because it will either return None, None or str, str

Not sure if this is the correct attitude, if type hints tell us that `None, str` might be returned, shouldn't we be prepared for handling it?
This kind of assumption seems very bug-inducing :thinking:

nazywam: The refactor lgtm tho

msm-code: `Oh so the tostring methods were added just to stop the linter from complaining? smiling_imp` well actually yes (because it technically was a correct code, it was linter that was wrong).

```
if type hints tell us that None, str might be returned, shouldn't we be prepared for handling it?
```
I guess the type should be `Union[Tuple[str, str], Tuple[None, None]` in the previous case :thinking:

msm-code: I mean, I agree that we shouldn't assume more than types say using knowledge about function internals, so good catch.

src/daemon.py
msm-code: :thinking: You're probably right (also the database doesn't *have* to be redis in the future, its an implementation defaul at this point). It's just, all the good names are already taken. I'll think of something and refactor.

src/db.py
```diff
@@ -0,0 +1,252 @@
...
+        return [cls(name) for name in names]
+
+    def __repr__(self) -> str:
+        return f"queue:{self.name}"
+
+
+class MatchInfo:
+    """ Represents information about a single match """
+
+    def __init__(self, file: str, meta: Dict[str, Any], matches: List[str]):
```
nazywam, line 64:
```suggestion
    def __init__(self, file: str, meta: Dict[str, Any], matches: List[str]) -> None:
```

PR #109, opened 2020-04-17
Initial implementation of Monaco text editor

### LINKED ISSUE #25: Add syntax highlighting to the Yara editor
**Description:**

The current Yara editor displaying the text black over white. It would be nicer to have it colorful with syntax highlighting applied. A list of Yara syntax-higlighters for different IDEs can be found here: https://github.com/InQuest/awesome-yara#syntax-highlighters

image

### LINKED ISSUE #26: Add tab key support to the Yara editor
**Description**
When pressing <kbd>TAB</kbd> in the Yara editor the website will navigate to the next HTML element by shifting the focus. When coding, <kbd>TAB</kbd> is frequently used for indention. Yara rules are rich with indentions and thus I believe it would be better that pressing on <kbd>TAB</kbd> will be treated as a `\t` in the editor.

Examples:
https://jsfiddle.net/2wAzx/13/
https://philipnewcomer.net/2015/11/how-to-make-the-tab-key-indent-text-in-a-texarea/
https://stackoverflow.com/questions/6637341/use-tab-to-indent-in-textarea

### LINKED ISSUE #27: Show line numbers in the Yara editor
**Description**
Line numbers are a useful feature in text editors. The current textarea does nto show line numbers. This is especially a problem when trying to compile a rule and getting and error saying that there is a problem in line 1234. A user would have to go to another IDE to see which line is the problematic

image

ITAYC0HEN:

The current "editor" is a simple text area. While it is nice for sending messages to our parents, it's not enough to write code, or in our case - yara rules.

This pull-request implements the Monaco text editor in mquery. this including all the built-in features of Monaco editor such as syntax highlighting, code completion, snippets, search, mini-map view, and more.

In addition, I modified the editor to work nice with Yara:
- Added the yara-language for enhanced syntax highlighting (TODO: extract it into an external OS package for the use of the world)
- Added `rule` snippet to ease the writing 
- Created a visible indicator for errors reported by yaramod when parsing

**Showoff + demo**
Overall look, syntax highlighting and code snippets:

mquery_rule_snippet
---

Parse and display errors:

mquery_pr_error

---

Read-only is a little bit grayed out, and shows a tool-tip when trying to paste data:

mquery_pr_readonly

---

**Left to do:**
- [x]  Fix a bug where the editor is emptied after executing a query
- [x] Add a border-primary around the editor. Cat knows why it doesn't work.
- [ ] Refine yara-lang and test against edge cases and crazy rules
- [ ] Optional: Color variable names
- [ ] Optional: send PR to `yaramod` to return a more verbose error struct, with errored lines.

**Test plan**

TODO

**Closing issues**

closes #25 , closes #26 , closes #27


ITAYC0HEN:
I know nothing about react, and although this is still a WIP, I feel like I really need more eyes here before I put on the extra work. Would love feedback from Reactors here ->> @msm-code @psrok1 @icedevml

msm-code:
Again, thanks for your PR!

```
[ ] Optional: send PR to yaramod to return a more verbose error struct, with errored lines.
```

That's a good idea. But even if we do this, it'll take a while before it gets merged, and new yaramod version is released - so i say let's leave it for now (and maybe add a tracking issue?)

src/mqueryfront/src/QueryField.js
```diff
@@ -31,18 +32,34 @@ class QueryField extends Component {
...
-            isLocked: newProps.isLocked,
-        });
+    static getDerivedStateFromProps(nextProps, prevState) {
+        if (
+            nextProps.rawYara !== prevState.rawYara ||
+            nextProps.readOnly !== prevState.readOnly
+        ) {
+            return { rawYara: nextProps.rawYara, readOnly: nextProps.readOnly };
+        }
+    }
```
msm-code, line 42: Nice - I knew we're doing it in an obsolete way, but never was motivated enough to refactor it.

src/mqueryfront/src/QueryMonaco.js
```diff
@@ -0,0 +1,162 @@
...
+                                        .InsertAsSnippet,
+                                documentation: "Generate a rule skeleton",
+                            },
+                        ];
+                        return { suggestions: suggestions };
+                    },
+                });
+
+                // We want to have a wa to indicate the user that the
+                // Editor is currently in read-only mode
```
msm-code, line 99:
```suggestion
                // We want to have a way to indicate the user that the
                // editor is currently in read-only mode
```

ITAYC0HEN:
> Small nitpick, I think it shouldn't colour the whole line, but only offending string?

That's not a nickpic!! That was a regression somewhere. Will fix :)

src/mqueryfront/src/QueryMonaco.js
ITAYC0HEN: `Editor` is the class of the object. Otherwise I'd write "that **the editor**"

msm-code: I just saw the typo in a `way`, the `Editor` one was a nitpick (but right, it's correct like it is)

ITAYC0HEN:
Fixed:
- Adde soft gray border
- Fix regression reported by @msm-code regarding token highlighting.



image

src/mqueryfront/src/QueryMonaco.js
```diff
@@ -0,0 +1,163 @@
...
+        }
+    }
+
+    setReadOnly(readOnly) {
+        if (this.state.editor) {
+            this.setState({ readOnly: readOnly });
+        }
+    }
+
+    setError(error) {
```
psrok1, line 35: Maybe we should unpack error to several arguments instead of treating it like unlabeled array. This will make setError more generic and ready for further changes.
```suggestion
    setError(error, startLine, startColumn, endColumn, errorMessage) {
```
We can call it using spread syntax in the regex part. Maybe we should include the actual errorMessage as well in case we want to show some tooltips?
```javascript
this.refs.editor.setError(
    ...err.match(/Error at (\d+).(\d+)-?(\d+)?: (.*)/)
);
```

src/mqueryfront/src/QueryMonaco.js
```diff
@@ -0,0 +1,163 @@
...
+                    [
+                        {
+                            range: new monaco.Range(
+                                error[1],
+                                error[2],
+                                error[1],
+                                Number(error[3]) + 1 // Increase by one because it excludes end column
+                            ),
+                            options: {
+                                isWholeLine: error.length > 3 ? false : true,
```
psrok1, line 54: According to the previous suggestion:
```suggestion
                                isWholeLine: endColumn !== undefined ? false : true,
```

src/mqueryfront/src/QueryField.js
```diff
@@ -206,14 +217,8 @@ class QueryField extends Component {
...
-                <div className="form-group">
-                    <textarea
-                        name="rawYara"
-                        className="form-control mquery-yara-input"
-                        onChange={this.handleInputChange}
-                        readOnly={this.state.isLocked}
-                        value={this.state.rawYara}
-                    />
+                <div className="mt-1 monaco-container">
+                    <QueryMonaco ref="editor" />
```
psrok1, line 221: String Refs are legacy (https://reactjs.org/docs/refs-and-the-dom.html#legacy-api-string-refs). Maybe we should use `React.createRef()`?

src/mqueryfront/src/QueryMonaco.js
```diff
@@ -0,0 +1,163 @@
+import React, { Component } from "react";
+import Editor, { monaco } from "@monaco-editor/react";
+import YARA from "./yara-lang";
+
+class QueryMonaco extends Component {
+    constructor(props) {
+        super(props);
+        this.state = {
+            rawYara: props.rawYara,
```
psrok1, line 9: Shouldn't we use `getDerivedStateFromProps` here? (https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops) It could be glitchy in case rawYara passed as property has been changed without remounting the component.

src/mqueryfront/src/QueryMonaco.js
ITAYC0HEN: omg I have no idea why I used if ternary here instead of just 
```
isWholeLine: error.length != 4
```
or in the new suggestion -> 
```
isWholeLine: encColumn == undefined
```

src/mqueryfront/src/QueryMonaco.js
```diff
@@ -0,0 +1,163 @@
...
+        }
+    }
+
+    setError(error) {
+        if (!this.state.editor || !error) {
+            return;
+        }
+
+        monaco.init().then((monaco) => {
+            this.state.decorations = this.state.editor
```
psrok1, line 41: Do not modify state directly, use `setState()`

https://reactjs.org/docs/state-and-lifecycle.html#do-not-modify-state-directly

src/mqueryfront/src/QueryMonaco.js
```diff
@@ -0,0 +1,163 @@
...
+                );
+        });
+    }
+
+    handleEditorDidMount(_, editor) {
+        this.setState({ editor, editor });
+
+        monaco
+            .init()
+            .then((monaco) => {
```
psrok1, line 69: Shouldn't it be initialized once e.g. on the top of file? I think we should separate monaco configuration from the component rendering.

https://github.com/suren-atoyan/monaco-react#monaco-instance

src/mqueryfront/src/QueryMonaco.js
psrok1: If you don't want to cause component re-rendering: maybe "decorations" is not a state and it should be placed in class attribute.

src/mqueryfront/src/QueryMonaco.js
```diff
@@ -0,0 +1,163 @@
...
+                height="70vh"
+                language="yara"
+                theme={this.state.readOnly ? "readOnlyTheme" : "light"}
+                value={this.state.rawYara}
+                editorDidMount={this.handleEditorDidMount}
+                options={{
+                    selectOnLineNumber: true,
+                    lineNumbersMinChars: 0,
+                    glyphMargin: true,
+                    readOnly: this.state.readOnly,
```
psrok1, line 152: I think that readOnly should be passed directly from the props and not to be part of state. Avoid using methods like `setReadonly`

icedevml:
@ITAYC0HEN If you are still willing to work on the PR, I think it should remain open :)

ITAYC0HEN:
wait wat ><" ... come one GtiHub... 

sorry folks, I commented using <kbd>Ctrl</kbd>+<kbd>Enter</kbd> and it closed  + commented instead of just comment....

Thanks @icedevml, it was a mistake and no intentional :)

src/mqueryfront/src/QueryMonaco.js
ITAYC0HEN: This suggestion contradicts the suggestions you mentioned on your second review. Thus, I feel like this one can be put aside ^

psrok1: That's true, it should be closed :)

ITAYC0HEN:
Fixed and improved most of the things :) I mean, at least those I understood.

Known issues:
- [ ] When refreshing (<kbd>F5</kbd>) the page when on a task (e.g `http://localhost/query/W4QKUGZ65MB9`) then the Editor is read-only but for some reason I can't understand, the read-only theme isn't applied to it.
- [ ] Some times, I noticed that the editor's width is smaller than it should be. Like it was created smaller than the usual. This one is hard to reproduce.

src/mqueryfront/src/QueryMonaco.js
psrok1: After some experiments: it seems that this is the only way to get the real "monaco" object. Accepted :)

src/mqueryfront/src/QueryField.js
```diff
@@ -1,17 +1,20 @@
 import React, { Component } from "react";
 import axios from "axios";
 import { API_URL } from "./config";
+import QueryMonaco from "./QueryMonaco";
 
 class QueryField extends Component {
     constructor(props) {
         super(props);
+        this.editor = React.createRef();
```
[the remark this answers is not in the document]
ITAYC0HEN:
```suggestion
```

PR #119, opened 2020-04-22
Change daemons to agents, to make distributed mquery possible

msm-code:
Big architecture change, even though it doesn't look like it.
Use agents instead of global daemons. Agents don't have to share filesystem
with other agents - they care only about their assigned ursadb instance.

This change is pretty experimental and there are some crucial UI improvements
that need to be done (for example, agent management from the web UI). But
looks like it works, so I hope to get it merged.

This PR also removes:
 - code that never worked well (db compacting, it's much more complicated than
   single button)
 - code obsoleted by the new concept (storage)
 - code that doesn't work yet anyway (metadata plugin integration) - will be restored soon after
 - temporarily job priority is ignored - will be restored soon after

- [ ] I've added automated tests for my change (if applicable, optional) - not yet

see above

see above.

- Instead of global queues, all tasks are stored in `agent:<agent_id>:queue-xxx` queues
- Redis keeps track of active agents
- New search jobs are automatically routed to active agents.

**Test plan**
Do some queries and ensure they work like they worked before

**Closing issues**


src/daemon.py
```diff
@@ -161,58 +68,88 @@ def execute_yara(job: JobId, files: List[str]) -> None:
...
-        db.finish_job(job)
 
+    db.job_update_work(job, len(files), num_matches)
 
-def execute_search(job_id: JobId) -> None:
-    logging.info("Parsing...")
 
-    job = db.get_job(job_id)
-    yara_rule = job.raw_yara
+def do_search(job: JobId, iterator: str):
```
BonusPlay, line 87:
```suggestion
def do_search(job: JobId, iterator: str) -> bool:
```

src/daemon.py
```diff
@@ -161,58 +68,88 @@ def execute_yara(job: JobId, files: List[str]) -> None:
...
+        job = JobId(data["job"])
+        iterator = data["iterator"]
+        if job.key in finished_jobs:
+            return
+        db.agent_start_job(agent_id, job, iterator)
+        if do_search(job, iterator):
+            finished_jobs.add(job.key)
+            db.agent_finish_job(agent_id, job)
+    else:
+        raise RuntimeError("Unsupported quue")
```
psrok1, line 135:
```suggestion
        raise RuntimeError("Unsupported queue")
```

src/daemon.py
```diff
@@ -161,58 +68,88 @@ def execute_yara(job: JobId, files: List[str]) -> None:
...
+        if pop_result.should_drop_iterator:
+            logging.info(
+                "Dropping job %s because iterator %s is empty",
+                job.key,
+                iterator,
+            )
+            return True
+    except Exception:
+        logging.exception("Failed to execute yara match.")
+        return True
```
psrok1, line 107: I think that logic is a bit unusual here: maybe we should return False in case of failure instead of True?

src/db.py
```diff
@@ -198,13 +101,34 @@ def job_contains(self, job: JobId, ordinal: str, file_path: str) -> bool:
         file_list = self.redis.lrange(job.meta_key, ordinal, ordinal)
         return file_list and file_path == json.loads(file_list[0])["file"]
 
+    def job_start_work(self, job: JobId, files_in_progress: int) -> None:
+        """ Updates the numbre of files being processed right now.
```
psrok1, line 105:
```suggestion
        """ Updates the number of files being processed right now.
```

src/daemon.py
psrok1: *in case of finish

src/db.py
```diff
@@ -215,57 +143,69 @@ def create_search_task(
...
+    def update_job_files(self, job: JobId, total_files: int) -> None:
+        self.redis.hincrby(job.key, "total_files", total_files)
+
+    def agent_start_job(
+        self, agent_id: str, job: JobId, iterator: str
+    ) -> None:
+        job_data = json.dumps({"job": job.key, "iterator": iterator})
+        self.redis.rpush(f"agent:{agent_id}:queue-yara", job_data)
+
+    def agent_finish_job(self, agent_id: str, job: JobId) -> None:
```
psrok1, line 202: `agent_id` is unreferenced

src/daemon.py
```diff
@@ -1,22 +1,32 @@
...
-from typing import Any, Dict, List
-from db import Database, JobId, MatchInfo
+from typing import Any, List, Dict
+from lib.yaraparse import parse_yara, combine_rules
+from db import AgentTask, JobId, Database, MatchInfo
+
 
-db = Database()
-ursa = UrsaDb(config.BACKEND)
+YARA_CACHE: Dict[str, Any] = {}
```
psrok1, line 14: I think it should be flushed somehow to prevent memory leak (seems to be a bit tricky).

Issue #113, opened 2020-04-20
Count number of samples in progress


msm-code:
**Feature Category**

- [x] Correctness
- [x] User Interface / User Experience
- [ ] Performance
- [ ] Other (please explain)

**Describe the problem**

Right now we don't track the number of samples that are being processed. This, combined with batching of samples, causes some non-obvious results in the UI, and makes it hard to know when the job is really done.

**Describe the solution you'd like**

Count numbre of samples being processed at any given moment.

Show them on the progress bar:

![ss](https://user-images.githubusercontent.com/7026881/79769065-f0806480-832b-11ea-809c-e9396943eed3.png)

Use them to determine when the job is done in the backend.

**Describe alternatives you've considered**

Hack around with more ifs in the frontend - not sustainable long term.

PR #120, opened 2020-04-22
add % of processing files to progress bar

KrzyWie:

Displaying the number of files in progres as a percentage of processing files from all files in progress bar.

**Test plan**
Run the query.

**Closing issues**
#113


src/mqueryfront/src/QueryResultsStatus.js
```diff
@@ -142,6 +142,10 @@ class QueryResultsStatus extends Component {
         let progress = Math.floor(
             (this.props.job.files_processed * 100) / this.props.job.total_files
         );
+        let processing = Math.floor(
+            (this.props.job.files_in_progress * 100) /
+                this.props.job.total_files
```
msm-code, line 147: what if total_files == 0?

src/mqueryfront/src/QueryResultsStatus.js
```diff
@@ -142,6 +142,10 @@ class QueryResultsStatus extends Component {
         let progress = Math.floor(
             (this.props.job.files_processed * 100) / this.props.job.total_files
         );
+        let processing = Math.floor(
```
msm-code, line 145: maybe round instead of floor

src/mqueryfront/src/QueryResultsStatus.js
```diff
@@ -226,6 +230,13 @@ class QueryResultsStatus extends Component {
                     >
                         {progress}%
                     </div>
+                    <div
+                        className={"progress-bar warning"}
+                        role="progressbar"
+                        style={{ width: processing + "%" }}
+                    >
+                        {processing}%
```
msm-code, line 238: what if processing% < 1? (it's rounded down to integer, so even 0.99% will show as 0%).

If processing is equal to 0% we probably shouldn't show this progress bar at all?

Also maybe consider rounding the processing to 2 decimal digits instead of to integers?

Also if the processing % is very small, it'll be too thin (narrower than the text) and it looks ugly. Maybe consider setting a minimum value for display width? (for example minimum 3%, so if `processsing > 0 && processing < 3`, use 3% for width).

src/mqueryfront/src/QueryResultsStatus.js
```diff
@@ -142,8 +142,16 @@ class QueryResultsStatus extends Component {
...
         );
+        let processing = Math.round(
+            (this.props.job.files_in_progress * 100) /
+                this.props.job.total_files
+        );
         let processed =
             this.props.job.files_processed + " / " + this.props.job.total_files;
+
+        if (processing > 0 && processing < 3) {
+            processing = 3;
```
msm-code, line 153: This will show `3%` processing even if there's only one file being processed. Comment about minimum with was about style only.

src/mqueryfront/src/QueryResultsStatus.js
```diff
@@ -226,6 +234,15 @@ class QueryResultsStatus extends Component {
                     >
                         {progress}%
                     </div>
+                    {processing > 0 && (
```
msm-code, line 237: total_files > 0?

src/mqueryfront/src/QueryResultsStatus.js
msm-code: (I suggest to remove this if and use max() in `style` below)

KrzyWie: Sorry, I understood the last review in a weird way and I was confused when implementing it.
New commit has condition that processing > 2 and total_files > 0

PR #121, opened 2020-04-22
Add a script for local indexing

msm-code:
Doing a lot of logic in ursadb `index` command didn't turn out well. Lift it
to python to allow for more parallelism in an user friendly way.

In the future, cache indexed files on disk to decrease RAM usage.

Requires `nocheck` support in ursadb.

- [n/a] I've tested my changes by building and running mquery, and testing changed functionality (if applicable)
- [n/a?] I've added automated tests for my change (if applicable, optional)
- [ ] I've updated documentation to reflect my change (if applicable) - will update later, when this becomes the recommended method

Indexing files does not parallelise well and uses a lot of RAM

Indexing is easy to scale to many processes, and RAM usage can be optimised later too.

**Test plan**

I've tested like this

```
python3 src/reindex_local.py samples/ --ursadb tcp://ursadb:9281 --tmpdir /usr/src/app/index/ --tmpdir-mount /var/lib/ursadb --path-mount /mnt/samples --batch 1 --workers 3
```

There are a lot of parameters here, because of the filesystem split. In case of bare metal ursadb deployment this would reduce to:

```
python3 src/reindex_local samples [plus optionally --batch and --workers]
```

**Closing issues**

### DIFF
--- a/src/reindex_local.py
+++ b/src/reindex_local.py
@@ -6,8 +6,8 @@ import argparse
 from typing import Set, Iterator, Tuple, List
 from pathlib import Path
 from lib.ursadb import UrsaDb
-from tempfile import NamedTemporaryFile
 from multiprocessing import Pool
+from tempfile import NamedTemporaryFile
 
 
 def all_indexed_files(ursa: UrsaDb) -> Set[str]:
--- a/src/reindex_local.py
+++ b/src/reindex_local.py
@@ -15,7 +15,7 @@ def all_indexed_files(ursa: UrsaDb) -> Set[str]:
     result: Set[str] = set()
     while True:
         pop_result = ursa.pop(iterator, 5000)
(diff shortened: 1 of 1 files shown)


src/reindex_local.py
```diff
@@ -0,0 +1,131 @@
...
+    for tmppath in tmpfiles:
+        mounted_name = os.path.join(
+            tmpdir_mount, os.path.relpath(tmppath, args.tmpdir)
+        )
+        indexing_jobs.append((args.ursadb, mounted_name))
+        logging.info(f" - Next batch: %s", mounted_name)
+
+    logging.info("Stage 3: run index command in parallel")
+    pool = Pool(processes=args.workers)
+    pool.map(index_files, indexing_jobs)
```
chivay, line 123: :angry_snake_noises:

PR #122, opened 2020-04-23
Make the /status page work with multiple agents.

msm-code:
Up until now the frontend assumed that there is only one ursadb instance.
Now every agent will register its ursadb instance, and frontend will
try to connect to it.

- [n/a?] I've updated documentation to reflect my change (if applicable)

See above - frontend assumes that there is only one ursadb.

ss

I've also went ahead and added two small tooltips:

ss

ss

**Test plan**

Ideally run mquery with two agents and observe that it works correctly

**Closing issues**

not any, but there should be :thinking: 

fixes #issuenumber

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -189,7 +189,7 @@ def backend_status_datasets() -> BackendStatusDatasetsSchema:
         except Again:
             pass
 
-    return BackendStatusDatasetsSchema(db_alive=True, datasets=datasets)
+    return BackendStatusDatasetsSchema(datasets=datasets)
 
 
 @app.get("/query/{path}", include_in_schema=False)
--- a/src/app.py
+++ b/src/app.py
@@ -1,6 +1,8 @@
+from lib.ursadb import UrsaDb
 import os
 
 import uvicorn
+import config
 from fastapi import FastAPI, Body, Query, HTTPException
 from starlette.requests import Request
 from starlette.responses import Response, FileResponse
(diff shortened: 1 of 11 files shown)


src/mqueryfront/src/VersionStatus.js
```diff
@@ -2,7 +2,14 @@ import React, { Component } from "react";
 
 class VersionStatus extends Component {
     render() {
-        const VersionRows = Object(this.props.components);
+        console.log(this.props.components);
```
psrok1, line 5: `console.log` s need to be removed.

src/mqueryfront/src/DatabaseTopology.js
```diff
@@ -85,15 +85,32 @@ class DatabaseTopology extends Component {
...
-                                <th>Dataset ID</th>
-                                <th> # Files (size)</th>
+                                <th
+                                    data-toggle="tooltip"
+                                    title={datasetTooltip}
+                                >
+                                    Dataset ID
+                                </th>
+                                <th data-toggle="tooltip" title={filesTooltip}>
+                                    {" "}
```
psrok1, line 111: Probably left by prettify - do we need leading space here?

src/app.py
```diff
@@ -157,39 +156,40 @@ def job_statuses() -> JobsSchema:
...
-    return BackendStatusDatasetsSchema(db_alive=db_alive, datasets=datasets)
+    datasets = {}
+    for url in db.get_active_agents().values():
+        try:
+            ursa = UrsaDb(url)
+            datasets.update(ursa.topology()["result"]["datasets"])
+        except Again:
+            pass
+
+    return BackendStatusDatasetsSchema(db_alive=True, datasets=datasets)
```
psrok1, line 192: `db_alive` is set to True even if there are no active agents - is it correct? Maybe that Schema field should be removed if it's no longer needed.

psrok1: Don't need to be stable if it's undocumented :smarts:

PR #124, opened 2020-04-23
Ignore env

msm-code:
Ignore .env file


msm-code:
Ok, I forgot that we can't remove .env, because it's used for default values :thinking:

PR #127, opened 2020-04-27
Fix/implement ursaexpression literal

### LINKED ISSUE #99: Implement UrsaExpression.literal
**Feature Category**

- [ ] Correctness
- [ ] Performance
- [x] Other (refactoring)

**Describe the problem**

Right now every time we want to match a literal in lib/yaraparse.py we need to write:

```python
UrsaExpression(f"{{{some_string.hex()}}}")
``` 

This is pretty repetitive and ugly

Implement `UrsaExpression.literal` and change all occurences of this pattern to:

```python
UrsaExpression.literal(some_string)
``` 

Dealing with ugly code forever

dskwhitehat:

f"{}" need to be added every time.

Logic stays unchanged. The only difference is that it's just sufficient to give pure string in literal function instead of adding "f{}".

**Test plan**
Run tests from src/tests

**Closing issues**

fixes #99

### DIFF
--- a/src/lib/yaraparse.py
+++ b/src/lib/yaraparse.py
@@ -2,7 +2,6 @@ import argparse
 import itertools
 import re
 from typing import Dict, List, Optional
-
 from yaramod import (  # type: ignore
     AndExpression,
     EqExpression,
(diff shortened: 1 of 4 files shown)


src/lib/yaraparse.py
```diff
@@ -136,15 +143,15 @@ def ursify_hex(hex_str: str) -> UrsaExpression:
             output.append(part[last_end:])
 
     core = "} & {".join(output)
-    return UrsaExpression(f"{{{core}}}")
+    return UrsaExpression.literal(core)
```
msm-code, line 146: It'll work, but it's a bit against the spirit of the idea (`core` is something like `616161} & {626262} & {636363`). Ideally I'd like to get rid of the default constructor. Maybe something like:

```
return UrsaExpression.and_(UrsaExpression.literal(f) for f in output)
```

dskwhitehat: Done

PR #128, opened 2020-04-27
Add simply duration and ETA to query status

KrzyWie:

Addition of time counting: E.T.A and Duration

**Test plan**
Run some querys

### DIFF
--- a/src/mqueryfront/src/QueryResultsStatus.js
+++ b/src/mqueryfront/src/QueryResultsStatus.js
@@ -3,6 +3,7 @@ import axios from "axios/index";
 import { API_URL } from "./config";
 import Pagination from "react-js-pagination";
 import QueryTimer from "./QueryTimer";
+import { finishedStatuses } from "./QueryUtils";
 
 function MatchItem(props) {
     const metadata = Object.values(props.meta).map((v) => (
@@ -174,7 +175,6 @@ class QueryResultsStatus extends Component {
 
         let progressBg = "bg-" + this.renderSwitchStatus(this.props.job.status);
 
-        const finishedStatuses = ["done", "cancelled", "failed", "expired"];
         if (finishedStatuses.includes(this.props.job.status)) {
             cancel = <span />;
         }
--- a/src/mqueryfront/src/QueryResultsStatus.js
+++ b/src/mqueryfront/src/QueryResultsStatus.js
@@ -261,16 +261,14 @@ class QueryResultsStatus extends Component {
                     <div className="col-md-3">
                         Processed: <span>{processed}</span>
                     </div>
-                    <div className="col-md-3">
-                        <span className="pull-right">
-                            <QueryTimer
-                                job={this.props.job}
-                                finishStatus={finishedStatuses}
-                                duration={true}
-                                eta={true}
-                            />{" "}
-                            {cancel}
-                        </span>
+                    <div className="col-md-3" style={{ "text-align": "right" }}>
+                        <QueryTimer
+                            job={this.props.job}
+                            finishStatus={finishedStatuses}
+                            duration={true}
+                            eta={true}
+                        />{" "}
+                        {cancel}
                     </div>
                 </div>
                 {this.props.job.error ? (
--- a/src/mqueryfront/src/QueryResultsStatus.js
+++ b/src/mqueryfront/src/QueryResultsStatus.js
@@ -174,7 +174,7 @@ class QueryResultsStatus extends Component {
 
         let progressBg = "bg-" + this.renderSwitchStatus(this.props.job.status);
 
-        let finishedStatuses = ["done", "cancelled", "expired"];
+        const finishedStatuses = ["done", "cancelled", "failed", "expired"];
         if (finishedStatuses.includes(this.props.job.status)) {
             cancel = <span />;
         }
@@ -265,6 +265,7 @@ class QueryResultsStatus extends Component {
                         <span className="pull-right">
                             <QueryTimer
                                 job={this.props.job}
+                                finishStatus={finishedStatuses}
                                 duration={true}
                                 eta={true}
                             />{" "}
--- a/src/mqueryfront/src/QueryResultsStatus.js
+++ b/src/mqueryfront/src/QueryResultsStatus.js
@@ -266,7 +266,7 @@ class QueryResultsStatus extends Component {
                             job={this.props.job}
                             finishStatus={finishedStatuses}
                             duration={true}
-                            eta={true}
+                            countDown={true}
                         />{" "}
                         {cancel}
                     </div>
--- a/src/mqueryfront/src/QueryResultsStatus.js
+++ b/src/mqueryfront/src/QueryResultsStatus.js
@@ -79,8 +79,6 @@ class QueryResultsStatus extends Component {
         this.state = {
             activePage: 1,
             itemsPerPage: 20,
-            // setTimer: true,
-            // countDown: 0
         };
 
         this.handleCancelJob = this.handleCancelJob.bind(this);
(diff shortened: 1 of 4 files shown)


src/mqueryfront/src/QueryTimer.js
```diff
@@ -0,0 +1,99 @@
...
+
+    render() {
+        let durationTime;
+        let clock = <span />;
+        let finishedStatuses = ["done", "cancelled", "failed", "expired"];
+        if (
+            this.props.job.submitted &&
+            !finishedStatuses.includes(this.props.job.status)
+        ) {
+            if (this.props.duration) {
```
msm-code, line 76: nicpick: unnecesary if, you can just always do the subtraction

src/mqueryfront/src/QueryTimer.js
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    render() {
+        let durationTime;
+        let clock = <span />;
+        let finishedStatuses = ["done", "cancelled", "failed", "expired"];
+        if (
+            this.props.job.submitted &&
+            !finishedStatuses.includes(this.props.job.status)
+        ) {
```
msm-code, line 75: To reduce nesting:

```suggestion
        if (
            !this.props.job.submitted ||
            finishedStatuses.includes(this.props.job.status)
        ) { return <span />; }
```
(and remove leftover braces)

msm-code: Or maybe remove this if from here, and instead don't add QueryTimer if the job is done?

src/mqueryfront/src/QueryTimer.js
```diff
@@ -0,0 +1,99 @@
...
+                    this.setState({ countDownTime: countDown });
+                }
+            }
+        }
+    }
+
+    render() {
+        let durationTime;
+        let clock = <span />;
+        let finishedStatuses = ["done", "cancelled", "failed", "expired"];
```
msm-code, line 71: duplicated code with `QueryResultsStatus.js`, probably move to some common file?

src/mqueryfront/src/QueryTimer.js
```diff
@@ -0,0 +1,99 @@
...
+                    processedFiles * processedTime - processedTime
+                );
+                if (this.state.firstCountDown) {
+                    this.setState({
+                        countDownTime: countDown,
+                        firstCountDown: false,
+                    });
+                } else if (
+                    this.state.countDownTime > countDown &&
+                    ["processing", "querying"].includes(this.props.job.status)
```
msm-code, line 60: check with finished statuses and negate the result instead

src/mqueryfront/src/QueryTimer.js
```diff
@@ -0,0 +1,99 @@
...
+        ) {
+            if (this.props.job.files_processed > 0 && this.props.eta) {
+                let processedFiles =
+                    this.props.job.total_files / this.props.job.files_processed;
+                let processedTime =
+                    this.state.currentTime - this.props.job.submitted;
+                let countDown = Math.round(
+                    processedFiles * processedTime - processedTime
+                );
+                if (this.state.firstCountDown) {
```
msm-code, line 53: why `firstCountDown`? :thinking:

src/mqueryfront/src/QueryTimer.js
```diff
@@ -0,0 +1,99 @@
...
+                    this.setState({
+                        countDownTime: countDown,
```
msm-code, line 55: I think you should compute `contDown` method in the `render` method instead (state should be avoided in react when possible)

src/mqueryfront/src/QueryTimer.js
```diff
@@ -0,0 +1,99 @@
...
+            this.setState({
+                currentTime: 0,
+                countDownTime: undefined,
+                firstCountDown: true,
+            });
+        }
+    }
+
+    setCountDownTime() {
+        let finishedStatuses = ["done", "cancelled", "failed", "expired"];
```
msm-code, line 40: duplicated code with `QueryResultsStatus.js` and this file, probably move to some common file?

src/mqueryfront/src/QueryResultsStatus.js
```diff
@@ -78,6 +79,8 @@ class QueryResultsStatus extends Component {
         this.state = {
             activePage: 1,
             itemsPerPage: 20,
+            // setTimer: true,
```
msm-code, line 82: (commented out code)

src/mqueryfront/src/QueryTimer.js
```diff
@@ -0,0 +1,99 @@
...
+            }
+
+            if (this.props.duration && this.props.eta) {
+                clock = (
+                    <i>
+                        {durationTime}s (~{this.state.countDownTime}s left)
+                    </i>
+                );
+            } else if (this.props.duration && !this.props.eta) {
+                clock = <i>{durationTime}s</i>;
```
psrok1, line 88: You can just return that component instead of putting it in the variable. This `<span>` wrapping is unnecessary in most cases.

src/mqueryfront/src/QueryTimer.js
```diff
@@ -0,0 +1,99 @@
...
+            } else if (!this.props.duration && this.props.eta) {
+                clock = <i>~{this.state.countDownTime}s</i>;
+            }
+        } else {
+            clock = <span />;
```
psrok1, line 93: If you don't want to place any components, you can just put empty array here.  It's equivalent of `no components to render`

psrok1: Even better: https://reactjs.org/docs/jsx-in-depth.html#booleans-null-and-undefined-are-ignored

src/mqueryfront/src/QueryTimer.js
```diff
@@ -0,0 +1,99 @@
...
+        this.setState({
+            currentTime: Math.floor(Date.now() / 1000),
+        });
+    }
+
+    componentDidMount() {
+        this.interval = setInterval(() => this.tick(), 1000);
+
+        if (this.props.eta) {
+            this.interwalCountDown = setInterval(
```
psrok1, line 19: Typo?
```suggestion
            this.intervalCountDown = setInterval(
```
and this interval is never cancelled. Why do we need two intervals instead of one when they are ticking with the same frequency?

src/mqueryfront/src/QueryTimer.js
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    componentWillUnmount() {
+        clearInterval(this.interval);
+    }
+
+    render() {
+        if (
+            !this.props.job.submitted ||
+            ["done", "cancelled", "failed", "expired"].includes(
```
msm-code, line 26: duplicated with `QueryPage.js` and `QueryResultsStatus.js` => move the array to some shared common file.

KrzyWie:
do we think about displaying time in a different way? E.T.A in minutes etc.
In my opinion this view looks good but converting seconds is a moment so it can be done quickly.
dureta

src/mqueryfront/src/QueryTimer.js
```diff
@@ -0,0 +1,106 @@
...
+                    {countdowmMin ? (
+                        <span>
+                            {countdowmMin}m {countDownTime}s
+                        </span>
+                    ) : (
+                        countDownTime >= 0 && <span>{countDownTime}s</span>
+                    )}
+                </i>
+            );
+        }
```
msm-code, line 103: Move the common code out of the ifs, and leave just:
```
if (this.props.duration && this.props.countDown) {
    return <i>{duration} <countdown>}</i>;
}
if (this.props.duration) {
    return duration;
}
if (this.props.countDown) {
    return countdown;
}

KrzyWie:
Screenshot from 2020-04-29 13-21-59

PR #131, opened 2020-04-29
Configurable agent plugins

### LINKED ISSUE #67: Rethink plugin system and support
Currently we have a small set of hardcoded metadata plugins, that we recently merged with the main repo.

This system is not extensible and hard to configure, even in our environment. Think how can we improve this and open it a bit.

psrok1:

- Plugins report desired configuration keys with description
- Agents report which plugins are successfully loaded and what is desired configuration schema
- Configuration is stored in database, so metadata plugins are configurable via UI
- Added `ExampleTagPlugin` to test plugin machinery, refactored existing plugins

**Related todo with this feature (in other PR)**
- Declaring plugins that are required to launch for current query (needed by integration)
- Respect additional plugin requirements

Related with #67

### DIFF
--- a/src/daemon.py
+++ b/src/daemon.py
@@ -10,7 +10,7 @@ from typing import Any, List
 from lib.yaraparse import parse_yara, combine_rules
 from db import AgentTask, JobId, Database, MatchInfo
 from cachetools import cached, LRUCache
-from metadata import MetadataPlugin
+from metadata import MetadataPlugin, Metadata
 from plugins import METADATA_PLUGINS
 
 
@@ -117,7 +117,7 @@ class Agent:
     ) -> None:
         """ After finding a match, push it into a database and
         update the related metadata """
-        metadata = {}
+        metadata: Metadata = {}
         for plugin in self.active_plugins:
             extracted_meta = plugin.run(file_path, metadata)
             metadata.update(extracted_meta)
--- a/src/daemon.py
+++ b/src/daemon.py
@@ -119,8 +119,13 @@ class Agent:
         update the related metadata """
         metadata: Metadata = {}
         for plugin in self.active_plugins:
-            extracted_meta = plugin.run(file_path, metadata)
-            metadata.update(extracted_meta)
+            try:
+                extracted_meta = plugin.run(file_path, metadata)
+                metadata.update(extracted_meta)
+            except Exception:
+                logging.exception(
+                    f"Failed to launch plugin {plugin.get_name()} for {file_path}"
+                )
         match = MatchInfo(file_path, metadata, matches)
         self.db.add_match(job, match)
 
--- a/src/daemon.py
+++ b/src/daemon.py
@@ -119,9 +119,6 @@ class Agent:
     ) -> None:
         """ After finding a match, push it into a database and
         update the related metadata """
-        if self.db.get_plugin_config_version() != self.plugin_config_version:
-            logging.info("Configuration changed - reloading plugins.")
-            self.__initialize_agent()
         metadata: Metadata = {}
         for plugin in self.active_plugins:
             try:
@@ -137,6 +134,9 @@ class Agent:
         self.db.add_match(job, match)
 
     def __execute_yara(self, job: JobId, files: List[str]) -> None:
+        if self.db.get_plugin_config_version() != self.plugin_config_version:
+            logging.info("Configuration changed - reloading plugins.")
+            self.__initialize_agent()
         rule = compile_yara(self.db, job)
         num_matches = 0
         self.db.job_start_work(job, len(files))
--- a/src/daemon.py
+++ b/src/daemon.py
@@ -129,7 +129,9 @@ class Agent:
                 metadata.update(extracted_meta)
             except Exception:
                 logging.exception(
-                    "Failed to launch plugin %s for %s", plugin.get_name(), file_path
+                    "Failed to launch plugin %s for %s",
+                    plugin.get_name(),
+                    file_path,
                 )
         match = MatchInfo(file_path, metadata, matches)
         self.db.add_match(job, match)
--- a/src/daemon.py
+++ b/src/daemon.py
@@ -10,6 +10,7 @@ from typing import Any, List
 from lib.yaraparse import parse_yara, combine_rules
 from db import AgentTask, JobId, Database, MatchInfo
 from cachetools import cached, LRUCache
+from metadata import MetadataPlugin
 from plugins import METADATA_PLUGINS
 
 
@@ -55,7 +56,7 @@ class Agent:
         self.ursa_url = ursa_url
         self.db = db
         self.ursa = UrsaDb(self.ursa_url)
-        self.active_plugins = []
+        self.active_plugins: List[MetadataPlugin] = []
 
     def __search_task(self, job_id: JobId) -> None:
         """Do ursadb query for yara belonging to the provided job.
@@ -89,13 +90,21 @@ class Agent:
             plugin_name = plugin_class.get_name()
             plugin_config = self.db.get_plugin_configuration(plugin_name)
             try:
-                active_plugins.append(
-                    plugin_class(self.db, plugin_config)
-                )
+                active_plugins.append(plugin_class(self.db, plugin_config))
                 logging.info(f"Loaded {plugin_name} plugin")
             except Exception:
                 logging.exception(f"Failed to load {plugin_name} plugin")
         self.active_plugins = active_plugins
+
+    def __initialize_agent(self) -> None:
+        self.__load_plugins()
+        plugin_spec = {
+            plugin.get_name(): plugin.__config_fields__
+            for plugin in self.active_plugins
+        }
+        self.db.register_active_agent(
+            self.group_id, self.ursa_url, plugin_spec
+        )
 
     def __update_metadata(
         self, job: JobId, file_path: str, matches: List[str]
@@ -185,6 +194,9 @@ class Agent:
                 logging.exception("Failed to execute task.")
                 self.db.fail_job(job, str(e))
                 self.db.agent_finish_job(job)
+        elif task.type == "reload":
+            # Reload plugins and reset agent registration
+            self.__initialize_agent()
         else:
             raise RuntimeError("Unsupported queue")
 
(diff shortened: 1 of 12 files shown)


src/db.py
```diff
@@ -173,6 +177,7 @@ def agent_get_task(self, agent_id: str) -> AgentTask:
         task_queues = [
             f"{agent_prefix}:queue-search",
             f"{agent_prefix}:queue-yara",
+            f"{agent_prefix}:queue-reload",
```
msm-code, line 180: move this to the top (order matters, when mquery is busy with yara matches it won't even try to reload config)

psrok1: Fixed

psrok1:
Follow-up from conversation with @msm-code:

That reload job doesn't work great, because we could have many agents with the same identity (belong to the same group) and there is one reload job per group.

TODO in this PR: Switch to configuration versioning by timestamp and check on "yara" job whether config changed and plugins need to be reloaded.

src/daemon.py
```diff
@@ -81,12 +84,49 @@ def __search_task(self, job_id: JobId) -> None:
...
         self.db.agent_start_job(self.group_id, job_id, iterator)
 
+    def __load_plugins(self) -> None:
+        active_plugins = []
+        for plugin_class in METADATA_PLUGINS:
+            plugin_name = plugin_class.get_name()
+            plugin_config = self.db.get_plugin_configuration(plugin_name)
+            try:
+                active_plugins.append(plugin_class(self.db, plugin_config))
+                logging.info(f"Loaded {plugin_name} plugin")
```
msm-code, line 94: nitpick but can't resist
```suggestion
                logging.info("Loaded %s plugin", plugin_name)
```

src/daemon.py
```diff
@@ -81,12 +84,49 @@ def __search_task(self, job_id: JobId) -> None:
...
+            except Exception:
+                logging.exception(f"Failed to load {plugin_name} plugin")
```
msm-code, line 96:
```suggestion
                logging.exception("Failed to load %s plugin", plugin_name)
```

src/daemon.py
```diff
@@ -81,12 +84,49 @@ def __search_task(self, job_id: JobId) -> None:
...
         update the related metadata """
-        match = MatchInfo(file_path, {}, matches)
+        metadata: Metadata = {}
+        for plugin in self.active_plugins:
+            try:
+                extracted_meta = plugin.run(file_path, metadata)
+                metadata.update(extracted_meta)
+            except Exception:
+                logging.exception(
+                    f"Failed to launch plugin {plugin.get_name()} for {file_path}"
```
msm-code, line 127:
```suggestion
                    "Failed to launch plugin %s for %s", plugin.get_name(), file_path
```

src/db.py
```diff
@@ -199,8 +207,73 @@ def agent_finish_job(self, job: JobId) -> None:
...
+        )
+
+    def get_active_agents(self) -> Dict[str, AgentSpecSchema]:
+        return {
+            name: AgentSpecSchema.parse_raw(spec)
+            for name, spec in self.redis.hgetall("agents").items()
+        }
+
+    def get_plugins_config(self) -> List[ConfigSchema]:
+        config_fields: Dict[str, Dict[str, str]] = {}
```
msm-code, line 234: maybe use defaultdict to make it a bit simpler?

src/db.py
```diff
@@ -199,8 +207,73 @@ def agent_finish_job(self, job: JobId) -> None:
...
+        # Merge all config fields
+        for agent_spec in self.get_active_agents().values():
+            for name, fields in agent_spec.plugins_spec.items():
+                if name not in config_fields:
+                    config_fields[name] = {}
+                config_fields[name].update(fields)
+        # Transform fields into ConfigSchema
```
msm-code, line 241: Uhh, now I see what you meant when you said it's `a spaghetti` (because of our data format)

src/db.py
```diff
@@ -199,8 +207,73 @@ def agent_finish_job(self, job: JobId) -> None:
...
+
+    def cache_get(self, key: str, expire: int) -> Optional[str]:
+        value = self.redis.get(f"cached:{key}")
+        if value is not None:
+            self.redis.expire(f"cached:{key}", expire)
+        return value
 
-    def get_active_agents(self) -> Dict[str, str]:
-        return self.redis.hgetall("agents")
+    def cache_store(self, key: str, value: str, expire: int):
```
msm-code, line 278:
```suggestion
    def cache_store(self, key: str, value: str, expire: int) -> None:
```

src/db.py
```diff
@@ -199,8 +207,73 @@ def agent_finish_job(self, job: JobId) -> None:
...
+            for plugin in sorted(plugin_configs.keys())
+            for key in sorted(plugin_configs[plugin].keys())
+        ]
+
+    def get_plugin_configuration(self, plugin_name: str) -> Dict[str, str]:
+        return self.redis.hgetall(f"plugin:{plugin_name}")
+
+    def set_plugin_configuration_key(
+        self, plugin_name: str, key: str, value: str
+    ):
```
msm-code, line 269:
```suggestion
    ) -> None:
```

src/db.py
```diff
@@ -158,6 +158,10 @@ def create_search_task(
             self.redis.rpush(f"agent:{agent}:queue-search", job.hash)
         return job
 
+    def create_reload_task(self, agents: List[str]):
```
msm-code, line 161:
```suggestion
    def create_reload_task(self, agents: List[str]) -> None:
```

src/metadata.py
```diff
@@ -1,50 +1,52 @@
...
 
 class MetadataPlugin(ABC):
-    #: List of plugin identifiers that plugin depends on
-    __depends_on__: List[str] = []
     #: Enables cache for extracted metadata
     __cacheable__: bool = False
     #: Overrides default cache expire time
     __cache_expire_time__: int = DEFAULT_CACHE_EXPIRE_TIME
+    #: Configuration keys required by plugin with description as a value
+    __config_fields__: Dict[str, str] = {}
```
msm-code, line 18:
```suggestion
    CONFIG_FIELDS: Dict[str, str] = {}
```

I probably should've complained earlier, but I don't think we should use names with underscores on both sides, because they're reserved.

https://dbader.org/blog/meaning-of-underscores-in-python

```
However, names that have both leading and trailing double underscores are reserved for special use in the language. This rule covers things like __init__ for object constructors, or __call__ to make an object callable.

These dunder methods are often referred to as magic methods—but many people in the Python community, including myself, don’t like that.

It’s best to stay away from using names that start and end with double underscores (“dunders”) in your own programs to avoid collisions with future changes to the Python language.
```

Also, to quote something more normative, PEP8:

```
__double_leading_and_trailing_underscore__: "magic" objects or attributes that live in user-controlled namespaces. E.g. __init__, __import__ or __file__. Never invent such names; only use them as documented.
```

(note that forward underscores only will apply name mangling, so they may not be what is intended here)

src/plugins/cuckoo_analysis.py
```diff
@@ -3,15 +3,17 @@
...
 from typing import Optional
 
-from metadata import Metadata, MetadataPlugin
+from db import Database
+from metadata import Metadata, MetadataPlugin, MetadataPluginConfig
 
 
 class CuckooAnalysisMetadata(MetadataPlugin):
     __cacheable__ = True
+    __config_fields__ = {"path": "Path to Cuckoo analyses"}
```
msm-code, line 12:
```suggestion
    __config_fields__ = {"path": "Root of cuckoo analysis directory."}
```

src/plugins/mwdb_uploads.py
```diff
@@ -5,24 +5,24 @@
...
     __cacheable__ = True
-
-    def __init__(
-        self, mwdb_api_token: str, mwdb_api_url: str, mwdb_url: str
-    ) -> None:
-        super().__init__()
-        self.mwdb = Malwarecage(api_url=mwdb_api_url, api_key=mwdb_api_token)
-        self.mwdb_url = mwdb_url
+    __config_fields__ = {
+        "mwdb_url": "Malwarecage URL accessible for mquery users",
```
msm-code, line 15:
```suggestion
        "mwdb_url": "URL to the malwarecage instance (for example, http://mwdb.cert.pl)",
```

src/plugins/mwdb_uploads.py
```diff
@@ -5,24 +5,24 @@
...
+        "mwdb_api_url": "API URL accessible from mquery daemon",
```
msm-code, line 16:
```suggestion
        "mwdb_api_url": "API URL to malwarecage instance (usually mwdb_url + '/api/')",
```

src/plugins/mwdb_uploads.py
```diff
@@ -5,24 +5,24 @@
...
+        "mwdb_api_token": "API key for 'mquery' user in Malwarecage",
```
msm-code, line 17:
```suggestion
        "mwdb_api_token": "API key for 'mquery' user in Malwarecage (base64encoded, starts with ey...)",
```

src/plugins/mwdb_uploads.py
msm-code: Also *technically* we could use a reasonable default here, but maybe it's not worth the effort.

src/metadata.py
psrok1: Yup, that was convention initiated before. It's good time to refactor that.

psrok1: But I think it should be still lowercase, it's rather class attribute than constant.

psrok1:
What about this version? 
- Configuration version check and reload is performed now by `Agent.__update_metadata`.
- Configuration change is detected by checking the `plugin-version` key timestamped by `self.redis.time()`. Anyway, it doesn't make a huge difference if we would just use `time.time()` from mquery-web, it's the only source of configuration setup.

src/daemon.py
```diff
@@ -81,12 +85,55 @@ def __search_task(self, job_id: JobId) -> None:
...
+            ],
+        )
+
     def __update_metadata(
         self, job: JobId, file_path: str, matches: List[str]
     ) -> None:
         """ After finding a match, push it into a database and
         update the related metadata """
-        match = MatchInfo(file_path, {}, matches)
+        if self.db.get_plugin_config_version() != self.plugin_config_version:
```
msm-code, line 122: Ugh, it's not a good place to do this - __update_metadata is in a realatively tight loop.

Move to a beginning of `__execute_yara` - in a worst case changes will be applied few seconds later, but we'll save a lot of requests to redis.

msm-code: (or to a beginning of `__yara_task`, that's good too)

psrok1: Moved to the beginning of `__execute_yara`

PR #134, opened 2020-04-30
Add various database debugging and management utilities

msm-code:
- [(waiting for release)] I've updated documentation to reflect my change (if applicable)

I have scripts that I use but they are only on my machine/servers

The scripts are shared with others

**Test plan**

Just run the scripts, they don't require any configuration.
 - nanobench will print some stats about time it took to execute various queries
 - compactall will try to compact the database, until it can't do it anymore.

### DIFF
--- a/src/utils/compactall.py
+++ b/src/utils/compactall.py
@@ -1,0 +1,43 @@
+import logging
+from lib.ursadb import UrsaDb
+import time
+import argparse
+
+
+def main() -> None:
+    logging.basicConfig(level=logging.INFO)
+
+    parser = argparse.ArgumentParser(description="Keep the database lean.")
+    parser.add_argument(
+        "--ursadb",
+        help="URL of the ursadb instance.",
+        default="tcp://localhost:9281",
+    )
+
+    args = parser.parse_args()
+    ursa = UrsaDb(args.ursadb)
+    stage = 0
+    last_datasets = None
+    while True:
+        datasets = set(
+            ursa.execute_command("topology;")["result"]["datasets"].keys()
+        )
+        if last_datasets:
+            removed = list(last_datasets - datasets)
+            created = list(datasets - last_datasets)
+            logging.info("%s => %s", removed, created)
+        logging.info("Stage %s: %s datasets left.", stage, len(datasets))
+        if last_datasets and datasets == last_datasets:
+            logging.info("Finally, a fixed point! Returning...")
+            return
+
+        start = time.time()
+        ursa.execute_command("compact all;")
+        end = time.time()
+        logging.info("Compacting took %s seconds...", (end - start))
+        stage += 1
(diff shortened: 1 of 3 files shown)


src/utils/compactall.py
```diff
@@ -0,0 +1,43 @@
+import logging
+from lib.ursadb import UrsaDb
+import time
+import argparse
+
+
+def main() -> None:
+    logging.basicConfig(level=logging.INFO)
+
+    parser = argparse.ArgumentParser(description="Keep the database lean.")
```
icedevml, line 10: Worth considering if these shouldn't be just UrsaDB feature/more advanced command

src/utils/nanobench.py
```diff
@@ -0,0 +1,129 @@
...
+    benchmark(socket, "select {61 62 63 64 ?? ?? 65};")
+    benchmark(socket, "select {65 ?? ?? 61 62 63 64};")
+    benchmark(socket, "select {?1 ?2 ?3 ?4 ?5 ?6};")
+    benchmark(socket, "select {61 62 63} & {?1 ?2 ?3 ?4 ?5 ?6};")
+    benchmark(socket, "select {61 62 63 ?? ?? ?? ?1 ?2 ?3 ?4 ?5 ?6 ?8};")
+    benchmark(socket, "select {?1 ?2 ?3 ?4 ?5 ?6} & {61 62 63};")
+    benchmark(socket, "select {?1 ?2 ?3 ?4 ?5 ?6 ?? ?? ?? 61 62 63};")
+
+
+def main() -> None:
```
icedevml, line 92: see `ursadb_bench` command in ursadb - this is similar functionality

msm-code: Not really though - ursadb_bench benchmarks `index+merge`, while this tests for query time.

But in the future I think `ursadb_bench` will get a rewrite in python too, just not now (e2e tests already do the same thing, but they don't measure time)

src/utils/compactall.py
msm-code: There is a benefit ofdoing this in another script: this script is NON-transactional. Ursadb tries hard to make every operation atomic. It's usually a nice property, but when compacting of a large database takes days, you don't want to lose all progress because of server reboot, for example.

See also the second script (reindex_local) that also has this non-property.

Of course, we could also make core ursadb non-transactional (it's very tempting), but so far I've resisted this temptation!

src/utils/nanobench.py
msm-code: Another advantage that I have in mind, that having this in an independent script will make it easy to benchmark all ursadb versions in a fair way - i.e. do some git magic to compile every ursadb version in existence, run benchmark on it and graph results. That sounds like a cool idea to see preformance improvements/regressions over the project's history.

PR #136, opened 2020-05-02
Support nocase expressions

### LINKED ISSUE #48: Support case insenstive strings in yara rules
Right now, we just ignore strings with the nocase flag:

```
rule CaseInsensitiveTextExample
{
    strings:
        $text_string = "foobar" nocase

    condition:
        $text_string
}
```

Supporting them correctly is... harder than it looks like. This can match:
```
foobar
foobaR
foobAr
foobAR
fooBar
... and 59 strings more
```

and ursadb query language is not expressive enough to support this.

We can't hack around this by chopping the query in the backend to something like:

```
( "foo" AND (
    "oob" AND (
        "oba" AND (
            ...
        ) OR
        "obA" AND (

        )
    ) OR
    "ooB" AND (
        "oBa" AND (
            ...
        ) OR
        "oBA" AND (
            ...
        )
    )
) OR "foO AND (
    ...
) OR "fOo" AND (
    ...
) OR "fOO" AND (
    ...
) ...
```

Because of exponential growth.

OTOH I feel like like this can solved with a C++ method (needs investigation). In this case we need to introduce `nocase` strings to ursadb.

Needs investigation (if this results in too many false positives, we may as well give up).

msm-code:

Nocase strings not supported

Nocase strings supported

**Test plan**
Run yara rule with nocase strings

**Closing issues**

fixes #48


nazywam:
image

According to this, strings with modifiers `nocase` and `wide` are legal, so we probably should think about supporting them

msm-code:
Good point!

I've added the support, and also the missing tests.

PR #145, opened 2020-05-03
Request plugin configuration to be reloaded on demand

psrok1:

- Agent checks whether configuration has changed during processing of Yara task. In "required plugins" feature (#144) we need to be sure that configuration is guaranteed to be updated by all agents and it will occur after relevant time.

After configuration change: reload task is sent to the `config-reload-{previous_version}` queue. Agent are listening for tasks related to the version they're currently on. When reload task is fetched: agent sends another request to be fetched by another agent and reloads the configuration. 

After 300 seconds of inactivity: mquery assumes that all agents have already abandoned that version and task expires.

**Test plan**

Tested manually:
- initial configuration set on single daemon
- configuration set with 3 agents sharing the same identity (`docker-compose -f docker-compose.dev.yml up --scale dev-daemon=3`)

**Closing issues**

Related with #67 
Dependency of #144

### DIFF
--- a/src/daemon.py
+++ b/src/daemon.py
@@ -154,9 +154,6 @@ class Agent:
         self.db.add_match(job, match)
 
     def __execute_yara(self, job: JobId, files: List[str]) -> None:
-        if self.db.get_plugin_config_version() != self.plugin_config_version:
-            logging.info("Configuration changed - reloading plugins.")
-            self.__initialize_agent()
         rule = compile_yara(self.db, job)
         num_matches = 0
         self.db.job_start_work(job, len(files))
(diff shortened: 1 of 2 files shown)


src/daemon.py
```diff
@@ -231,7 +227,23 @@ def __process_task(self, task: AgentTask) -> None:
         :type task: AgentTask
         :raises RuntimeError: Task with unsupported type given.
         """
-        if task.type == TaskType.SEARCH:
+        if task.type == TaskType.RELOAD:
+            if (
+                self.plugin_config_version
+                == self.db.get_plugin_config_version()
+            ):
+                # This should never happen and suggests that version is not updated properly.
```
msm-code, line 235: IMO runtime error here is too harsh:
 - 1) configuration is only increasing, so this shouldn't cause a catastrophic failure and even if it's wrong, configuration will fix itself in the next version
 - 2) this kills the daemon

Maybe just logging.error() and continue processing?

src/db.py
```diff
@@ -190,15 +191,27 @@ def get_job_matches(
...
-    def agent_get_task(self, agent_id: str) -> AgentTask:
+    def reload_configuration(self, config_version: str):
+        # Send request to any of agents that configuration must be reloaded
+        self.redis.lpush(f"config-reload:{config_version}", "reload")
+        # After 300 seconds of inactivity: reload request is deleted
+        self.redis.expire(f"config-reload:{config_version}", 300)
+
+    def agent_get_task(self, agent_id: str, config_version: str) -> AgentTask:
         agent_prefix = f"agent:{agent_id}"
+        # config-reload is notification queue that is set by web to notify
```
msm-code, line 202:
```suggestion
        # config-reload is a notification queue that is set by web to notify
```

src/daemon.py
msm-code: The bigger WTF would be config versoin in the db that's older than current :thinking:. So maybe add `>=` here instead of `==`

psrok1: Actually I was afraid of situation when reload task is waiting in queue and plugin_config_version in database is somehow equal to the current version. This situation should never happen, but can cause infinite loop. `logging.error` should be enough.

psrok1: Version is a string, so for `>=` check we need to convert it to int and replace `initial` with 0 🤔

Issue #101, opened 2020-04-16
Show number of errored samples


msm-code:
**Feature Category**

- [x] Correctness
- [x] User Interface / User Experience
- [ ] Performance
- [ ] Other (please explain)

**Describe the problem**

When yara rule matching fails (for example, file does not exist), we just silently ignore the match and don't inform the user about the problem.

**Describe the solution you'd like**

Show a number of errors in the UI and on the progress bar (in red)

**Describe alternatives you've considered**

Change query status to `errored` after every error? Probably too harsh.

KWMORALE:
![image](https://user-images.githubusercontent.com/63305564/80708556-2d96e480-8aec-11ea-9743-b5e5b35035ac.png)

PR #148, opened 2020-05-04
Show number of errored samples

KWMORALE:

backend information that the file was not found

Counting number of errored samples using a separate function which in the future may allow it to be placed in more places. Sending this information via redis to the front-end and displaying how many files was missed. The information in the backend still remained.

**Test plan**
index some files and then delete some of them next run some query

**Closing issues**

#101

### DIFF
--- a/src/daemon.py
+++ b/src/daemon.py
@@ -98,21 +98,20 @@ class Agent:
             try:
                 matches = rule.match(sample)
                 if matches:
-                    num_matches += 1
                     self.__update_metadata(
                         job, sample, [r.rule for r in matches]
                     )
+                    num_matches += 1
             except yara.Error:
-                logging.exception(f"Yara failed to check file {sample}")
+                logging.error("Yara failed to check file %s", sample)
                 num_errors += 1
             except FileNotFoundError:
-                logging.exception(
-                    f"Failed to open file for yara check: {sample}"
-                )
+                logging.error("Failed to open file for yara check: %s", sample)
                 num_errors += 1
-        logging.exception(f"Error files: {num_errors}")
 
-        self.db.job_update_error(job, num_errors)
+        if num_errors > 0:
+            self.db.job_update_error(job, num_errors)
+
         self.db.job_update_work(job, len(files), num_matches)
 
     def __yara_task(self, job: JobId, iterator: str) -> None:
--- a/src/daemon.py
+++ b/src/daemon.py
@@ -92,6 +92,7 @@ class Agent:
     def __execute_yara(self, job: JobId, files: List[str]) -> None:
         rule = compile_yara(self.db, job)
         num_matches = 0
+        num_errors = 0
         self.db.job_start_work(job, len(files))
         for sample in files:
             try:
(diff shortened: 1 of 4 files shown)


src/daemon.py
```diff
@@ -103,11 +104,15 @@ def __execute_yara(self, job: JobId, files: List[str]) -> None:
                     )
             except yara.Error:
                 logging.exception(f"Yara failed to check file {sample}")
+                num_errors += 1
             except FileNotFoundError:
                 logging.exception(
                     f"Failed to open file for yara check: {sample}"
                 )
+                num_errors += 1
+        logging.exception(f"Error files: {num_errors}")
```
msm-code, line 113: 1) This is not related to an exception, so `logging.error` at most
2) This is being logged unconditionally (there should be no errors in logs when there are no errors)
3) As a general rule, in case of logging, prefer `logging.func("asdf %s", param)`  instead of `logging.func(f"asdf {param}")`
4) finally, this log is unncessary - all errors are logged already with more details anyway 

```suggestion
        logging.exception(f"Error files: {num_errors}")
```

src/mqueryfront/src/QueryResultsStatus.js
```diff
@@ -220,7 +220,7 @@ class QueryResultsStatus extends Component {
                 </div>
             );
         }
-
+        console.log(this.props.job.files_error);
```
msm-code, line 223: unnecessary console log

src/daemon.py
```diff
@@ -103,11 +104,15 @@ def __execute_yara(self, job: JobId, files: List[str]) -> None:
                     )
```
msm-code, line 104: move `num_matches` after the `__update_metadata` call, in case it throws an exception

src/mqueryfront/src/QueryResultsStatus.js
```diff
@@ -257,6 +257,12 @@ class QueryResultsStatus extends Component {
                         >
                             {this.props.job.status}
                         </span>
+                        {this.props.job.total_files - lenMatches > 0 && (
```
msm-code, line 260: I'd prefer to have the number of errored samples on the progress bar (like the samples in progress, but in red).

msm-code:
(also, please correct the flake8 warning)

KWMORALE:
image
At the moment it looks like this.
In my opinion, it is worth somewhere to place information about how many files were not found not just at the progress bar, in this commit it is next to the status.

src/mqueryfront/src/QueryResultsStatus.js
```diff
@@ -257,6 +268,12 @@ class QueryResultsStatus extends Component {
                         >
                             {this.props.job.status}
                         </span>
+                        {this.props.job.files_errored > 0 && (
+                            <span style={{ color: "grey", fontSize: "12px" }}>
+                                {" "}
+                                {this.props.job.files_errored} files was missed
```
msm-code, line 274: `*were` missed.

Or maybe just something like `xxx errors encountered` / `xxx errors during processing`?

msm-code: Also, maybe move it to the right of processing instead?

Also, not sure why but I have this (0 files)

ss

msm-code: Nvm the screenshot, I had the older version.
Still think that maybe it should be moved to the right of `processed`, and maybe displayed in red instead of being gray?

KWMORALE: Red color looked bad there, gray nicely entered the UI.
I was also thinking about putting it there (the right of processed).

PR #150, opened 2020-05-06
Incrementing config version instead of using second-resolution timestamp

psrok1:

- If configuration fields are changed too fast (in single second), configuration version doesn't change twice, but two reload tasks are spawned, so agent reports a bug:

```
Critical error: Requested to reload configuration, but configuration present in database is still the same (<epoch>).
```

- Version is converted to `int` instead of `str` and `INCRBY` is used for getting new version instead of `TIME` with `unix time in seconds` part. Initial value is 0.

### DIFF
--- a/src/plugins/__init__.py
+++ b/src/plugins/__init__.py
@@ -1,5 +1,6 @@
 from typing import List, Type
 from metadata import MetadataPlugin
+from .example_plugin import ExampleTagPlugin
 
 # Feel free to import plugins here and add them to list below
-METADATA_PLUGINS: List[Type[MetadataPlugin]] = []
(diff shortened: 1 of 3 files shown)


src/plugins/__init__.py
```diff
@@ -1,5 +1,6 @@
 from typing import List, Type
 from metadata import MetadataPlugin
+from .example_plugin import ExampleTagPlugin
 
 # Feel free to import plugins here and add them to list below
-METADATA_PLUGINS: List[Type[MetadataPlugin]] = []
+METADATA_PLUGINS: List[Type[MetadataPlugin]] = [ExampleTagPlugin]
```
msm-code, line 6: is this intentional?

psrok1: Oops

PR #152, opened 2020-05-06
Feature/#22

### LINKED ISSUE #22: Filter previous jobs by author
**Description**
Currently, mqeury parses the rule's author from the metadata part in the rule.

When the number of jobs will start to rise, it would be harder to return to find a previous job. It's even more important if a big organization is using mquery with > 10 daily users.
Thus, it would be nice to have an option to filter the jobs by their author.

image

---

*In general, I believe that implementing user management would be a good decision. Maybe something that is based on mwdb user permission model. This will help to make mquery suitable for big organizations. For example, some queries are too sensitive to share even with other people in the same organization. In addition, the author signed on a Yara rule isn't necessarily the person who executes the yara. Say, if someone is executing a rule from the internet written by a stranger.  Later, it will be hard for them to find this job since they don't remember the name of the stranger.

I'll open a new issue for this ^*

JaroPowerH:
- [x ] I've read the [contributing guideline](https://github.com/CERT-Polska/mquery/blob/master/CONTRIBUTING.md).

I also moved state handling from SearchJobs and SearchJobRows to parent class (RecentPage). Other classes converted to stateless functional components.
Thus, separated logic (RecentPage) from view (SearchJobs & other components).

fixes #22
RecentJobsFiltering3


src/mqueryfront/src/RecentPage.js
```diff
@@ -1,17 +1,135 @@
...
+            jobs: [],
+            head: [],
+            filter: null,
+            error: null,
+        };
+
+        this.handleCancelJob = this.handleCancelJob.bind(this);
+        this.handleClose = this.handleClose.bind(this);
+        this.handleFilter = this.handleFilter.bind(this);
+        this._getHead = this._getHead.bind(this);
```
msm-code, line 21: Why underscored? AFAIK underscores in javascript are pretty polarising (i.e. not everyone likes them). Also I think they're not more private than handleClose for example. So maybe they should just be called `getHead` and `getDistinctList`?

I don't have a strong opinion here, just know that some JS style guides disallow underscores in field names

src/mqueryfront/src/RecentPage.js
```diff
@@ -1,17 +1,135 @@
...
+                attrubuteName: "",
+                valueList: null,
+            },
+        ];
+
+        return head;
+    }
+
+    _getDistinctList(arrayOfObjects, attributeName, allElement) {
+        let returnList;
```
msm-code, line 93: maybe
```suggestion
        let returnList = null;    
```
to avoid returning undefined

src/mqueryfront/src/SearchJobs.js
```diff
@@ -1,154 +1,129 @@
...
+    } = props.job;
+    const rule_author = props.job.rule_author
+        ? props.job.rule_author
+        : "(no author)";
+    const { onClose, onCancel } = props;
+
+    const submittedDate = new Date(submitted * 1000).toISOString();
+
+    const actionBtn =
+        status === "cancelled" || status === "expired" || status === "done" ? (
```
msm-code, line 30: Maybe import QueryUtils.js and check for

```suggestion
        finishedStatuses.includes(status)
```

src/mqueryfront/src/RecentPage.js
```diff
@@ -1,17 +1,135 @@
...
+
+        if (arrayOfObjects)
+            returnList = arrayOfObjects
+                .map((item) => item[attributeName])
+                .filter((value, index, self) => self.indexOf(value) === index)
+                .sort();
```
msm-code, line 99: Matches are sorted as strings instead of numerically. This looks a bit weird, do you have an idea how to fix it? :thinking: 

ss

src/mqueryfront/src/RecentPage.js
JaroPowerH: JS do not support private methods. I used pseudo privacy by marking "private" method with underscore. See: https://stackoverflow.com/questions/27849064/how-to-implement-private-method-in-es6-class-with-traceur

src/mqueryfront/src/SearchJobs.js
JaroPowerH: I agree. It's the best solution. I'l apply this change in next commit.
My code was based on original one:
"       if (this.props.status === "cancelled" || this.state.cancelled) {
            status = "cancelled";
            cancelBtn = <span />;
        }

        if (this.props.status === "expired") {
            cancelBtn = "";
        }

        if (this.props.status === "done") {
            cancelBtn = "";
        }
"

src/mqueryfront/src/RecentPage.js
JaroPowerH: Good remark. I didn't notice it. I will correct it in next commit. I will use .sort((a, b) => a - b) for Matches.

src/mqueryfront/src/RecentPage.js
msm-code: Yes, OTOH all methods here are technically private (`handleClose` is private too, because external components are not supposed to call it).

I ask because it's not very common in react code. Some style guides (like AirBNB) even explicitly disallow them: https://github.com/airbnb/javascript/tree/master/react#methods.

I'm not decided either way, but we don't use this convention anywhere else so not sure about this.

src/mqueryfront/src/RecentPage.js
```diff
@@ -1,17 +1,182 @@
...
 class RecentPage extends Component {
+    constructor(props) {
+        super(props);
+
+        this.state = {
+            jobs: [],
+            filter: null,
+            head: [],
+            activePage: 1,
+            itemsPerPage: 11,
```
msm-code, line 16: A bit unintuitive, I'd assume that pages have 10 elements

```suggestion
            itemsPerPage: 10,
```

src/mqueryfront/src/RecentPage.js
```diff
@@ -1,17 +1,182 @@
...
+        sortNumerically = false,
+        allElement = "All"
+    ) {
+        let returnList = null;
+
+        if (arrayOfObjects) {
+            returnList = arrayOfObjects
+                .map((item) => item[attributeName])
+                .filter((value, index, self) => self.indexOf(value) === index)
+                .sort();
```
msm-code, line 116: This sort is not needed now, right?

```suggestion
                .filter((value, index, self) => self.indexOf(value) === index);
```

src/mqueryfront/src/SearchJobs.js
```diff
@@ -1,154 +1,174 @@
...
+
+const SearchJobRowEmpty = () => {
+    return (
+        <tr>
+            <td>
+                <div className="d-flex">
+                    <span className="invisible">sth</span>
+                </div>
+                <p className="mb-0 invisible" style={{ fontSize: 11 }}>
+                    sth
```
msm-code, line 21: Looks like accident, I'd change to something else like:

```suggestion
                    &nbsp;
```

src/mqueryfront/src/SearchJobs.js
```diff
@@ -1,154 +1,174 @@
...
+import Pagination from "react-js-pagination";
+import PropTypes from "prop-types";
+import { finishedStatuses } from "./QueryUtils";
+
+const SearchJobRowEmpty = () => {
+    return (
+        <tr>
+            <td>
+                <div className="d-flex">
+                    <span className="invisible">sth</span>
```
msm-code, line 18: Looks like accident, I'd change to something else like:
```suggestion
                    <span className="invisible">&nbsp;</span>
```

src/mqueryfront/src/RecentPage.js
JaroPowerH: OK.

src/mqueryfront/src/RecentPage.js
JaroPowerH: Of course!

src/mqueryfront/src/SearchJobs.js
JaroPowerH: OK. Invisible character in invisible element. Why not?

src/mqueryfront/src/SearchJobs.js
JaroPowerH: OK.

Issue #34, opened 2020-04-03
Add a Copy button for for all hashes, and for each single hash


ITAYC0HEN:
**Description**
Currently, it is not easy to retrieve the hash of the sample. If you're lucky, the filename is the hash, and then you'll need to copy each filename.
I suggest to add buttons to:
- Copy single hash (might be implemented by clicking on the cell)
- Copy all hashes as
    - Sha256
    - Sha1
    - MD5

![image](https://user-images.githubusercontent.com/20182642/78386181-c289e900-75e5-11ea-9d1d-b7116a136c28.png)

![image](https://user-images.githubusercontent.com/20182642/78386240-e0574e00-75e5-11ea-91c7-079b32bb1079.png)

Related to #33

ITAYC0HEN:
Does not solve this completely because Copy All Hashes wasn't implemented

msm-code:
Right, I've misunderstood this issue and thought that `copy all hashes` means "download a file with all hashes" (done with #163).

Actually I didn't even know that there's an option to copy all hashes. Do you find it useful often? To implement it we'll have to compute all the hashes after the query for every file, or think about caching them somewhere.

I'd like to avoid having too much metadata in mquery (it's not useful for us, because we use mwdb for that), but I can reconsider and/or create a new issue for that if that's important.

ITAYC0HEN:
aren't the hashes there already when displaying the sha256 on the table? Or they are calculated for 20 samples at a time?

msm-code:
Nevermind, looks like I've missed the point twice in a row. I've actually checked VT how it works now,

Yeah, all sha256 hashes are in the db. I don't think we can reuse the endpoint used to download the list of the hashes, but the code will be almost the same. I'll reopen it.

Issue #16, opened 2020-01-23
Display hash values along with samples when results are displayed


3vangel1st:
Could you add the SHA1 value of the file when results are displayed? That way we can also remove the physical samples from the directory ones they are indexed.

icedevml:
@3vangel1st Unfortunately this is not that simple. You do still need the physical samples to be in place, because the indexing process is lossy. In general it is not possible to recover the samples back from the indexes.

Also, the database is designed to work very fast but sometimes it may yield a false positive result. Depending on the query and the contents of your samples, this may happen more or less often.

For that reason mquery takes the search result set and checks the original files with YARA in the final step just to be sure. The original file must be available at this particular moment.

/cc @msm-code

msm-code:
Hi,

I've sent you an email, but I'll copy the details here for bookkeeping purposes:

there are actually two related but distinct systems: mquery and ursadb.
 - ursadb is responsible for indexing files, and after the initial indexing it doesn't need the files - but still keeps the original filename in its metadata. Queries are very fast, but they can return some false-positives.
 - mquery uses ursadb under the hood for querying. When querying for files, it queries ursadb and then checks all the files with yara rule again, to weed out false-positives.

We *could* allow mquery users to skip the additional yara scan step. In this case, mquery may return false positives. If that's OK with you, ping me and I'll be happy to implement this feature in my available time.

OTOH, can you explain why you want to remove the samples after scanning? In our use case (and in the other instances that I know of) people usually just index their "malware NFS". If you store samples in a weird location (like s3 buckets or sth), you can try to mount it with FUSE for ursadb, or (if that's not possible) we can consider adding a plugin support for mquery (so that you'll be able to index files coming from whatever your backing malware storage is).

As for the other part of your question (adding SHA1) - yes, that's doable. Not right now, but it's not hard to implement and we can do it if you're interested. If you remove the physical samples, you'll still need to deal with false positives somehow. Is that OK with you?

What do you think?

ITAYC0HEN:
I suggest renaming this issue to reflect the content. It might be included in #33

msm-code:
There are two ways to approach this problem:

 - implement additional attributes stored along with samples in the backend (something that we considered, but don't want to do this right now)
 - just recalculate the hashes every time they are needed.

I've added tags assuming the second implementation - this means that the physical samples will still be needed after indexing though.

msm-code:
[...]

PR #156, opened 2020-05-07
Sha256 value for each matched file

KWMORALE:

Show sha256 hash along with matched samples.
Sha256 is calculated in daemon by external function for each sample file and then it is sent via metadata.

**Closing issues**
#34


src/mqueryfront/src/QueryResultsStatus.js
```diff
@@ -9,12 +9,24 @@ function MatchItem(props) {
...
+                <span className="badge badge-pill badge-warning">
+                    {v.display_text}
+                </span>
+            )}
         </a>
     ));
 
+    const hashes = Object.values(props.meta).map((v) => (
+        <span>
+            {v.hidden && (
```
msm-code, line 22: instead of relying on `hidden`, find `v` with key equal to `sha256`

src/mqueryfront/src/QueryResultsStatus.js
```diff
@@ -206,6 +219,7 @@ class QueryResultsStatus extends Component {
                         <thead>
                             <tr>
                                 <th>Matches</th>
+                                <th style={{ width: "32%" }}>SHA256</th>
```
msm-code, line 222: weirdly specific

Use bootstrap classes (probably something like `col-md-4`). Bootstrap grid system will probably also allow to automatically hide/flow the columns on resize

src/setsha.py
```diff
@@ -0,0 +1,13 @@
+import hashlib
+ 
+def set_sha(filename):
```
msm-code, line 3: add type annotations

KWMORALE:
image

few changes regarding display:
1. tags are displayed in the next row instead of shortening the sample name.
2. column with sha 256 hashes has been added, where after clicking on the hash is copied to the clipboard.
3. the hash column is hidden when the page resolution is reduced and in the editor view

src/mqueryfront/src/QueryResultsStatus.js
```diff
@@ -9,12 +9,24 @@ function MatchItem(props) {
     const metadata = Object.values(props.meta).map((v) => (
         <a href={v.url}>
             {" "}
-            <span className="badge badge-pill badge-warning">
-                {v.display_text}
-            </span>
+            {!v.hidden && (
```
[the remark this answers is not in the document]
msm-code: (or actually you should just filter it before the .map()  )

KWMORALE:
Closing issue: #16 
if we consider sha256 instead of sha1

msm-code:
Not really closing #16, because this feature won't work after removing physical samples after indexing. To do this we'd have to keep the hashes in ursadb somehow

PR #159, opened 2020-05-11
Add filter plugins

msm-code:
It's currently not easily possible to remove files from mquery. As a
workaround, users can create their own blacklists for files that
should be ignored (for example, using a list of files that
should be ignored, or regular expression for file types that are
not interesting).

This also potentially allows us to query some database to check for
additional constraints on files (for example, cross check with mwdb).

- [soon] I've updated documentation to reflect my change (if applicable) 

No way to prefilter ursadb results

Added code for prefiltering, and example plugin

**Test plan**
- Index some files
- Add a simple blacklist regex in the config, for example "^0" - this should exclude all files that start with 0
- Ensure that no files that start with 0 are in the result

### DIFF
--- a/src/config.docker.py
+++ b/src/config.docker.py
@@ -6,3 +6,4 @@ REDIS_PORT = int(os.environ.get("REDIS_PORT", 6379))
 JOB_EXPIRATION_MINUTES = int(
     os.environ.get("JOB_EXPIRATION_MINUTES", 0)
 )  # infinite by default
+PLUGINS = list(filter(None, os.environ.get("MQUERY_PLUGINS", "").split()))
--- a/src/config.docker.py
+++ b/src/config.docker.py
@@ -1,4 +1,5 @@
 import os
+from plugins import parse_plugin_list
 
 BACKEND = os.environ.get("MQUERY_BACKEND", "tcp://ursadb:9281")
 REDIS_HOST = os.environ.get("REDIS_HOST", "redis")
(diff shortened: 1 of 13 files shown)


src/metadata.py
```diff
@@ -72,20 +84,36 @@ def run(self, matched_fname: str, current_meta: Metadata) -> Metadata:
             self._cache_store(identifier, result)
         return result
 
-    @abstractmethod
+    def filter(self, matched_fname: str) -> Optional[str]:
+        """
+        Checks if the file is a good candidate for further processing,
+        and fix the file path if necessary.
+        :param matched_fname: Matched file path
+        :return: New path to file (may be the same path). None if the file
```
psrok1, line 92: What's the use case of this path substitution?

src/daemon.py
```diff
@@ -157,26 +159,47 @@ def __execute_yara(self, job: JobId, files: List[str]) -> None:
...
+            filemap = new_filemap
+
+        for orig_name, path in filemap.items():
             try:
-                matches = rule.match(sample)
+                matches = rule.match(path)
                 if matches:
                     self.__update_metadata(
-                        job, sample, [r.rule for r in matches]
+                        job, orig_name, [r.rule for r in matches]
```
psrok1, line 184: Not sure if we expect original name or real path here :thinking:

src/config.docker.py
```diff
@@ -6,3 +6,4 @@
 JOB_EXPIRATION_MINUTES = int(
     os.environ.get("JOB_EXPIRATION_MINUTES", 0)
 )  # infinite by default
+PLUGINS = list(filter(None, os.environ.get("MQUERY_PLUGINS", "").split()))
```
psrok1, line 9: maybe `,` is better separator than whitespace? 🤔 But then it needs strip() for `", "`

src/daemon.py
```diff
@@ -5,13 +5,16 @@
...
-from util import setup_logging, update_sha
+from util import setup_logging, make_sha256_tag
 from typing import Any, List
 from lib.yaraparse import parse_yara, combine_rules
 from db import AgentTask, JobId, Database, MatchInfo, TaskType
 from cachetools import cached, LRUCache
 from metadata import MetadataPlugin, Metadata
-from plugins import METADATA_PLUGINS
+from plugins import load_plugins
+
```
psrok1, line 15: Too many empty lines
```suggestion
```

PR #164, opened 2020-05-13
Feature/multiselect for sample tags

### LINKED ISSUE #139: Multiselect for sample tags
**Feature Category**

**Describe the problem**

It's not possible to select more than one tag for samples

Change the (ugly as hell, btw) tag select button:

ss

To a multiselect:

ss

It doesn't have to be pretty, but it should:

 - look similar to the rest of the UI (use boostrap)
 - integrate with react (so it should **not** be implemented with a random jquery plugin)
 - work :P.

This will also require a small change in the backend (change parameter type from Optional[str] to List[str] in a few places)

dskwhitehat:
- [x ] I've read the [contributing guideline](https://github.com/CERT-Polska/mquery/blob/master/CONTRIBUTING.md).

Only one taint option could be chosen at a time.

Multiselect was prepared to be able to choose multiple options.

**Test plan**
Index files and prepare taints from datasets. Select taints you want or leave "everywhere" option. 

**Closing issues**

fixes #139

### DIFF
--- a/src/db.py
+++ b/src/db.py
@@ -106,7 +106,7 @@ class Database:
             total_files=int(data.get("total_files", 0)),
             files_errored=int(data.get("files_errored", 0)),
             iterator=data.get("iterator", None),
-            taint=data.get("taint", None),
+            taints=json.loads(data.get("taints", "[]")),
             total_datasets=data.get("total_datasets", 0),
             datasets_left=data.get("datasets_left", 0),
         )
@@ -154,7 +154,7 @@ class Database:
         rule_author: str,
         raw_yara: str,
         priority: Optional[str],
-        taint: Optional[str],
+        taints: List[str],
         agents: List[str],
     ) -> JobId:
         job = JobId(
@@ -181,8 +181,8 @@ class Database:
             "datasets_left": 0,
             "total_datasets": 0,
         }
-        if taint is not None:
-            job_obj["taint"] = taint
+
+        job_obj["taints"] = json.dumps(taints)
 
         self.redis.hmset(job.key, job_obj)
         for agent in agents:
--- a/src/lib/ursadb.py
+++ b/src/lib/ursadb.py
@@ -43,16 +43,17 @@ class UrsaDb:
     def query(
         self,
         query: str,
-        taint: Optional[str] = None,
+        taints: List[str] = [],
         dataset: Optional[str] = None,
     ) -> Json:
         socket = self.make_socket(recv_timeout=-1)
 
         start = time.clock()
         command = "select "
-        if taint:
-            taint = taint.replace('"', '\\"')
-            command += f'with taints ["{taint}"] '
+        if taints:
+            taints_str = '", "'.join(taints)
+            taints_whole_str = '["' + taints_str + '"]'
+            command += f"with taints {taints_whole_str} "
         if dataset:
             command += f'with datasets ["{dataset}"] '
 
--- a/src/mqueryfront/src/QueryField.js
+++ b/src/mqueryfront/src/QueryField.js
@@ -11,8 +11,6 @@ class QueryField extends Component {
             };
         });
 
-        console.log("Options: "+options)
-
         var multiselect = null;
 
         if (this.props.availableTaints.length) {
--- a/src/mqueryfront/src/QueryField.js
+++ b/src/mqueryfront/src/QueryField.js
@@ -1,8 +1,30 @@
 import React, { Component } from "react";
 import QueryMonaco from "./QueryMonaco";
+import ReactMultiSelectCheckboxes from "react-multiselect-checkboxes";
 
 class QueryField extends Component {
     render() {
+        var options = this.props.availableTaints.map(function (obj) {
+            return {
+                label: obj,
+                value: obj,
+            };
+        });
+
+        console.log("Options: "+options)
+
+        var multiselect = null;
+
+        if (this.props.availableTaints.length) {
+            multiselect = (
+                <ReactMultiSelectCheckboxes
+                    onChange={this.props.handleChange}
+                    options={options}
+                    placeholderButtonLabel="everywhere"
+                />
+            );
+        }
+
         return (
             <div>
                 <div className="btn-group mb-1" role="group">
@@ -71,37 +93,7 @@ class QueryField extends Component {
                             <span className="fa fa-code" /> Parse
                         </button>
                     )}
-                    <div className="btn-group" role="group">
-                        <button
-                            type="button"
-                            className="btn btn-info dropdown-toggle"
-                            data-toggle="dropdown"
-                            aria-haspopup="true"
-                            aria-expanded="false"
-                        >
-                            Search: {this.props.selectedTaint || "everywhere"}
-                        </button>
-                        <div className="dropdown-menu">
-                            <button
-                                className="dropdown-item"
-                                onClick={() => this.props.selectTaint(null)}
-                            >
-                                everywhere
-                            </button>
-                            {this.props.availableTaints.map((taint) => {
-                                return (
-                                    <button
-                                        className="dropdown-item"
-                                        onClick={() =>
-                                            this.props.selectTaint(taint)
-                                        }
-                                    >
-                                        {taint}
-                                    </button>
-                                );
-                            })}
-                        </div>
-                    </div>
+                    {multiselect}
                 </div>
                 <div className="mt-1 monaco-container">
                     <QueryMonaco
--- a/src/schema.py
+++ b/src/schema.py
@@ -18,7 +18,7 @@ class JobSchema(BaseModel):
     total_files: int
     files_errored: int
     iterator: Optional[str]
-    taint: Optional[str]
+    taints: List[str]
     datasets_left: int
     total_datasets: int
 
@@ -56,7 +56,7 @@ class RequestConfigEdit(BaseModel):
 
 class QueryRequestSchema(BaseModel):
     raw_yara: str
-    taint: Optional[str]
+    taints: Optional[List[str]]
     priority: Optional[str]
     method: str
     required_plugins: List[str] = Field([])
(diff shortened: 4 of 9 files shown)


src/db.py
```diff
@@ -99,7 +100,7 @@ def get_job(self, job: JobId) -> JobSchema:
             total_files=int(data.get("total_files", 0)),
             files_errored=int(data.get("files_errored", 0)),
             iterator=data.get("iterator", None),
-            taint=data.get("taint", None),
+            taint=json.loads(data.get("taint", None)),
```
msm-code, line 103: change key name in json to `taints`

also this code doesn't make sense right now, because json.loads(None) is an error - change to "[]"

src/db.py
```diff
@@ -145,7 +146,7 @@ def create_search_task(
         rule_author: str,
         raw_yara: str,
         priority: Optional[str],
-        taint: Optional[str],
+        taint: List[str],
```
msm-code, line 149: plural

src/db.py
```diff
@@ -170,8 +171,9 @@ def create_search_task(
             "files_errored": 0,
             "agents_left": len(agents),
         }
+
         if taint is not None:
```
msm-code, line 175: `taints` are not nullable now, so there is no need for this check

src/lib/ursadb.py
```diff
@@ -43,16 +42,17 @@ def make_socket(self, recv_timeout: int = 2000) -> zmq.Context:
     def query(
         self,
         query: str,
-        taint: Optional[str] = None,
+        taint: List[str] = None,
```
msm-code, line 45: plural

src/lib/ursadb.py
```diff
@@ -43,16 +42,17 @@ def make_socket(self, recv_timeout: int = 2000) -> zmq.Context:
...
     ) -> Json:
         socket = self.make_socket(recv_timeout=-1)
 
         start = time.clock()
         command = "select "
         if taint:
-            taint = taint.replace('"', '\\"')
-            command += f'with taints ["{taint}"] '
+            taint_str = str(taint)
+            taint_str = taint_str.replace("'", '"')
```
msm-code, line 54: what if taint name contains `"` or `'`?

You should create them one by one and then `", ".join` instead of relying on `str()`

src/mqueryfront/src/QueryField.js
```diff
@@ -1,14 +1,16 @@
...
 import QueryMonaco from "./QueryMonaco";
 
 class QueryField extends Component {
     constructor(props) {
         super(props);
 
         this.state = {
-            selectedTaint: null,
+            rawYara: props.rawYara,
+            selectedOption: null,
```
msm-code, line 13: empty option list is `[]`, not `null`

src/mqueryfront/src/QueryField.js
```diff
@@ -1,14 +1,16 @@
...
+import ReactMultiSelectCheckboxes from "react-multiselect-checkboxes";
 import QueryMonaco from "./QueryMonaco";
 
 class QueryField extends Component {
     constructor(props) {
         super(props);
 
         this.state = {
-            selectedTaint: null,
+            rawYara: props.rawYara,
```
msm-code, line 12: is rawYara necessary here?

src/mqueryfront/src/QueryField.js
```diff
@@ -20,32 +22,33 @@ class QueryField extends Component {
...
+    };
 
     handleYaraChanged(value) {
         this.props.updateYara(value);
     }
 
     handleQuery(event, method, priority) {
+        var taints = [];
+
+        if (this.state.selectedOption != null) {
```
msm-code, line 36: you don't need to create this list when the default value for state is `[]` I think?

Even then, you can just `var taints = this.state.selectedOption || []`

src/mqueryfront/src/QueryField.js
msm-code: also; maybe `selectedTaints` instead of `selectedOption`?

src/mqueryfront/src/QueryField.js
```diff
@@ -91,6 +94,37 @@ class QueryField extends Component {
     }
 
     render() {
+        var options = [];
+        var i;
+        for (i = 0; i < this.props.availableTaints.length; i++) {
```
msm-code, line 99: use Array.map (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map) here instead

src/mqueryfront/src/QueryField.js
```diff
@@ -91,6 +94,37 @@ class QueryField extends Component {
...
+            options.push({
+                label: this.props.availableTaints[i],
+                value: this.props.availableTaints[i],
+            });
+        }
+
+        var multiselect = "";
```
msm-code, line 106: Avoid creating an empty span unnecessarily:
```suggestion
        var multiselect = null;
```

src/mqueryfront/src/QueryField.js
```diff
@@ -91,6 +94,37 @@ class QueryField extends Component {
...
+
+        if (this.props.availableTaints != "") {
```
msm-code, line 108: isn't availableTaints a list? why compare with empty string

src/schema.py
```diff
@@ -17,7 +17,7 @@ class JobSchema(BaseModel):
     total_files: int
     files_errored: int
     iterator: Optional[str]
-    taint: Optional[str]
+    taint: List[str]
```
msm-code, line 20: use plural version here

src/schema.py
```diff
@@ -53,7 +53,7 @@ class RequestConfigEdit(BaseModel):
 
 class QueryRequestSchema(BaseModel):
     raw_yara: str
-    taint: Optional[str]
+    taint: List[str]
```
msm-code, line 56: plural (taints)

src/mqueryfront/src/QueryField.js
dskwhitehat: I tried to change this but every time I tried sth different it failed. What is your suggestion (working) ?

msm-code: to be clear, did you try:
```javascript
if (this.props.availableTaints)
```
and it doesn't work?

dskwhitehat: No. If there are no taints multiselect is still displayed but without taints.

src/lib/ursadb.py
```diff
@@ -43,16 +42,17 @@ def make_socket(self, recv_timeout: int = 2000) -> zmq.Context:
...
 
         start = time.clock()
         command = "select "
-        if taint:
-            taint = taint.replace('"', '\\"')
-            command += f'with taints ["{taint}"] '
+        if taints:
+            taints_str = ", ".join(taints)
+            taints_whole_str = '["' + taints_str + '"]'
+            command += f"with taints {taints_whole_str} "
```
msm-code, line 55: this will only work for a single taint, and cause a syntax error when there is more than 1 tag.

msm-code: `["asd", "qwe", "zxc"] ->  ["asd, qwe, zxc"]`

src/mqueryfront/src/QueryField.js
psrok1: Empty list is also truthy in JavaScript. You need to check `if(this.props.availableTaints.length)`.

Psst... `[] == []` is also false :D

psrok1: but `[] == ""` is true and that's why current code is working

dskwhitehat: Ok change with "if(this.props.availableTaints.length)" works right now. Tested manually.

src/lib/ursadb.py
```diff
@@ -43,16 +42,17 @@ def make_socket(self, recv_timeout: int = 2000) -> zmq.Context:
     def query(
         self,
         query: str,
-        taint: Optional[str] = None,
+        taints: List[str] = None,
```
msm-code, line 45: default value is not of type `List[str]`

```suggestion
        taints: List[str] = [],
```

src/mqueryfront/src/QueryField.js
msm-code: (still relevant, doesn't look like it's used)

src/mqueryfront/src/QueryField.js
```diff
@@ -91,6 +87,36 @@ class QueryField extends Component {
     }
 
     render() {
+        var options = this.props.availableTaints.map(function (obj) {
+            let rObj = {
```
msm-code, line 91:
```
return {...}
```

instead of
```
let rObj = {...}
return rObj;
```

src/schema.py
```diff
@@ -18,7 +18,7 @@ class JobSchema(BaseModel):
     total_files: int
     files_errored: int
     iterator: Optional[str]
-    taint: Optional[str]
+    taints: List[str]
```
msm-code, line 21: (not sure if we can change the public api, give me a second)

msm-code: Or actually, change to Optional[List[str]] please.

msm-code: wait, wrong line. This is OK actually.

src/schema.py
```diff
@@ -56,7 +56,7 @@ class RequestConfigEdit(BaseModel):
 
 class QueryRequestSchema(BaseModel):
     raw_yara: str
-    taint: Optional[str]
+    taints: List[str]
```
msm-code, line 59: Change to Optional[List[str]]

msm-code:
Problem 1: the button disappears when the user clicks on the `query` or `logo` buttons. This should be probably fixed before merging?

bug

You might want to check merge with master and see if the problem still occurs (@psrok1 and @JaroPowerH worked on similar bugs here)

Problem 2: taints are not selected properly when the job is started, or  opened from the recent's job page. This can be fixed in a separate PR if you want to merge this first.
bug1

PR #166, opened 2020-05-13
Query utils

JaroPowerH:
- [ x] I've read the [contributing guideline](https://github.com/CERT-Polska/mquery/blob/master/CONTRIBUTING.md).

A bunch of cosmetic changes around QueryUtils with no visual effect to make code better.

### DIFF
--- a/src/mqueryfront/src/QueryResultsStatus.js
+++ b/src/mqueryfront/src/QueryResultsStatus.js
@@ -3,7 +3,11 @@ import axios from "axios/index";
 import { API_URL } from "./config";
 import Pagination from "react-js-pagination";
 import QueryTimer from "./QueryTimer";
-import { finishedStatuses } from "./QueryUtils";
+import {
+    isStatusFinished,
+    getProgressBarClass,
+    getBadgeClass,
+} from "./queryUtils";
 import { CopyToClipboard } from "react-copy-to-clipboard";
 import ActionCancel from "./components/ActionCancel";
 
@@ -117,22 +121,6 @@ class QueryResultsStatus extends Component {
         this.sendResultsActivePage(pageNumber);
     }
 
-    renderSwitchStatus(status) {
-        switch (status) {
-            case "done":
-                return "success";
-            case "cancelled":
-                return "danger";
-            case "expired":
-                return "warning";
-            case "processing":
-            case "querying":
-                return "info";
-            default:
-                return "info";
-        }
-    }
-
     componentDidUpdate(prevProps) {
         if (prevProps.qhash !== this.props.qhash) {
             this.setState({ activePage: 1 });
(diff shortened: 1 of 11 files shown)


src/mqueryfront/src/components/StatusProgress.js
```diff
@@ -1,35 +1,14 @@
...
+    const { status, total_files, files_processed } = props;
 
-    const progressClass = "progress-bar bg-" + rowClass;
+    const progressClass = getProgressBarClass(status);
 
     const percentage = Math.round(
-        props.total_files
-            ? (props.files_processed * 100) / props.total_files
-            : 0
+        total_files ? (files_processed * 100) / total_files : 0
```
msm-code, line 11: There's a small bug here btw, and maybe it's a good time for fix it:

when the status is finished (for example, `failed`, 0/0 should render as completely filled in).

For example, `done` job that matched 0 files should be 100% green instead of empty.
The same goes for `failed` job that errored during parsing.

So instead of `0`, maybe it should be something like `isStatusFinished(status) ? 100 : 0`

src/mqueryfront/src/queryUtils.js
```diff
@@ -0,0 +1,38 @@
...
+        case "processing":
+        case "querying":
+            classSufix = "info";
+            break;
+        case "expired":
+            classSufix = "warning";
+            break;
+        case "cancelled":
+            classSufix = "danger";
+            break;
```
msm-code, line 20: We have a new status now - `removed. Maybe use `dark` for it?

src/mqueryfront/src/queryUtils.js
```diff
@@ -0,0 +1,38 @@
...
+        default:
+            break;
```
msm-code, line 22: maybe add some console log to make future debugging easier here.

src/mqueryfront/src/queryUtils.js
```diff
@@ -0,0 +1,38 @@
+export const isStatusFinished = (status) =>
+    ["done", "cancelled", "failed", "expired"].includes(status);
```
msm-code, line 2: `"removed"` should also be added here

src/mqueryfront/src/QueryResultsStatus.js
```diff
@@ -190,20 +181,19 @@ class QueryResultsStatus extends Component {
...
-        let progressBg = "bg-" + this.renderSwitchStatus(this.props.job.status);
-
-        if (finishedStatuses.includes(this.props.job.status)) {
+        const isFinished = isStatusFinished(status);
+        if (isFinished) {
             cancel = <span />;
         }
 
-        const lenMatches = this.props.job.files_matched;
+        const lenMatches = files_matched;
```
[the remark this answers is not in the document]
JaroPowerH: I agree. files_matched as props value is read only and there are no need to use lenMatches.

src/mqueryfront/src/components/StatusProgress.js
JaroPowerH: OK. I pass isFinished flag by props to component. After this change, it looks better.

src/mqueryfront/src/queryUtils.js
JaroPowerH: "dark" seems reasonable. Works for background and badge (`bg-dark,` `badge-dark`).

src/mqueryfront/src/queryUtils.js
JaroPowerH: OK. Unknown status will be "visible" by lack of colouring. I added: 
console.log(`getClassForStatus: unknown status="${status}"`);

src/mqueryfront/src/queryUtils.js
JaroPowerH: OK.

PR #167, opened 2020-05-13
Small formatting for the results table

ITAYC0HEN:

Currently, the results table contains two columns, one for hashes and the other for sha256.
When clicking on the hash in a cell, the hash will be copied to the clipboard
image

Since we enjoy tons of space on this table, and because we plan to add more info (file size, mwdb data, ...) I thought of making a few cosmetic changes.

Changes made:
- One column to contain everything nicely
- I believe that the sha256 is more important for the user than the internal filename, so I made it the main item in the
- A copy icon for both file name and hashes, visible on cell hover
- A download icon, visible on cell hover

image

Peek 2020-05-13 20-43

**Test plan**
 - Check that hashes are copied correctly
 - Check that filenames are copied correctly
 - Check that download works correctly
 - Check files with loooooong file names (can use the browser's developers-tools for this)

### DIFF
--- a/src/mqueryfront/src/QueryResultsStatus.js
+++ b/src/mqueryfront/src/QueryResultsStatus.js
@@ -8,6 +8,17 @@ import { CopyToClipboard } from "react-copy-to-clipboard";
 import ActionCancel from "./components/ActionCancel";
 
 function MatchItem(props) {
+    const download_url =
+        API_URL +
+        "/download?job_id=" +
+        encodeURIComponent(props.qhash) +
+        "&ordinal=" +
+        encodeURIComponent(props.ordinal) +
+        "&file_path=" +
+        encodeURIComponent(props.file);
+
+    const path = require("path");
+
     const metadata = Object.values(props.meta)
         .filter((v) => !v.hidden)
         .map((v) => (
@@ -17,20 +28,6 @@ function MatchItem(props) {
                     {v.display_text}
                 </span>
             </a>
-        ));
-
-    let hashes = Object.values(props.meta)
-        .filter((v) => v.hidden)
-        .map((v) => (
-            <CopyToClipboard text={v.display_text} key={v}>
-                <div
-                    style={{ fontFamily: "monospace" }}
-                    data-toggle="tooltip"
-                    title="Click to copy"
-                >
-                    {v.display_text}
-                </div>
-            </CopyToClipboard>
         ));
 
     let matches = <span></span>;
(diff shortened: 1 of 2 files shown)


src/mqueryfront/src/QueryResultsStatus.js
```diff
@@ -44,35 +41,53 @@ function MatchItem(props) {
...
+                                    className="text-secondary"
+                                >
+                                    <i className="fa fa-download fa-xm ml-2" />
+                                </a>
+                                <CopyToClipboard
+                                    text={props.meta.sha256.display_text}
+                                >
+                                    <span
+                                        data-toggle="tooltip"
+                                        title="Copy sha256 to clipboard"
```
msm-code, line 65: I *think* it should be `Copy sha256 to the clipboard`, but both forms are used on the internet so correct me if I'm wrong.

```suggestion
                                        title="Copy sha256 to the clipboard"
```

src/mqueryfront/src/QueryResultsStatus.js
```diff
@@ -44,35 +41,53 @@ function MatchItem(props) {
...
+                                </CopyToClipboard>
+                            </small>
+                        </div>
+                        <small class="text-secondary">
                             {path.basename(props.file)}
-                        </a>
+                            <CopyToClipboard text={path.basename(props.file)}>
+                                <span
+                                    data-toggle="tooltip"
+                                    title="Copy file name to clipboard"
```
msm-code, line 78: Again, I *think* an article is missing here

```suggestion
                                    title="Copy file name to the clipboard"
```

src/mqueryfront/src/QueryResultsStatus.js
```diff
@@ -44,35 +41,53 @@ function MatchItem(props) {
...
+                                >
+                                    <i className="fa fa-download fa-xm ml-2" />
+                                </a>
+                                <CopyToClipboard
+                                    text={props.meta.sha256.display_text}
+                                >
+                                    <span
+                                        data-toggle="tooltip"
+                                        title="Copy sha256 to clipboard"
+                                        className="hash-item"
```
msm-code, line 66: Why is this class called `hash-item`? Shouldn't it be something along the lines of `copyable-item` (since it's reused for filename)?

msm-code: Also, maybe consider adding it to CopyToClipboard (since looks like the intention is that every item in CopyToClipboard should cause mouse pointer to change). Does it sound reasonable? :thinking:

src/mqueryfront/src/QueryResultsStatus.js
BonusPlay: Actually that would imply, that there could be multiple clipboards (so you need to specify which one to copy to). *Copy sha256 to clipboard* is perfectly fine.

src/mqueryfront/src/QueryResultsStatus.js
BonusPlay: Nope.

src/mqueryfront/src/QueryResultsStatus.js
msm-code: `Copy to clipboard` is popular enough on the internet that you're probably right, I was just a bit confused because of the number of counter-examples on the internet, for example:

- [mozilla docs](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Interact_with_the_clipboard) - `Writing to the clipboard`, `copying the current selection into the clipboard` etc
- [office support](https://support.office.com/en-us/article/use-the-office-clipboard-028903c7-66ea-4eb0-b0a1-bacdb952e3a1) - `Cut and paste items from the Clipboard`, `To paste everything in the Clipboard`
- [wikipedia](https://en.wikipedia.org/wiki/Clipboard_(computing)) -  all `clipboards` prefixed with `the`
- [random tech site](https://www.digitalcitizen.life/simple-questions-what-clipboard-windows-how-manage-it) - `copy to the clipboard overwrites the previous contents of the clipboard`
- [w3schools](https://www.w3schools.com/howto/howto_js_copy_clipboard.asp) is bipolar on this, because post title is `How TO - Copy Text to Clipboard` and the description is `Learn how to copy text to the clipboard with JavaScript`.

Intuitively your explaination makes sense though, so i guess nevermind my comment here :+1:

BonusPlay: Well, english is hard. For example you have *Sun goes around **the** world*. You don't use *the* to specify which world, but there's only 1 (and it's obvious). However, if the context isn't as obvious (as clipboard) skipping *the* can be used to imply that there's only one clipboard. 

My guess that both versions are correct. In theory you should add *the*, but by omitting you provide more information to user.

ITAYC0HEN: Thanks, you two <3

PR #169, opened 2020-05-14
Add partial regex string support

### LINKED ISSUE #97: Investigate regex support
**Feature Category**

- [x] Correctness
- [x] Performance
- [ ] Other (please explain)

**Describe the problem**

We completely ignore any regexes in yara rules

After https://github.com/CERT-Polska/ursadb/pull/56 is implemented, we'll be able to use much more powerful matching engine in ursadb.

We should handle regexes where we can (for example `/abcd(e|f)gh/` should be easy to handle with a proper regex parser).

Keep ignoring regexes - accuracy suffers.

nazywam:

Regex strings are ignored when converting yara rules to ursa queries

Some regex strings should be correctly converted to ursa expressions

**Test plan**
run `lib/yaraparse.py` on yara files that do and don't contain regex strings

**Closing issues**

fixes #97


src/lib/yaraparse.py
```diff
@@ -193,13 +293,17 @@ def ursify_string(string: String) -> Optional[UrsaExpression]:
...
+            is_wide=string.is_wide,
+            is_nocase=string.is_nocase,
+        )
     elif string.is_hex:
         value_safe = string.pure_text.decode()
         return ursify_hex(value_safe)
     elif string.is_regexp:
-        # Not supported at this moment
-        return None
+        return ursify_regex_string(string)
```
msm-code, line 306: :proud_duck:

src/lib/yaraparse.py
```diff
@@ -153,20 +159,114 @@ def ursify_nocase_bytes(raw: bytes) -> UrsaExpression:
...
+        right = flatten_regex_or_tree(unit.right)
+        if not left or not right:
+            return None
+        return left + right
+    elif type(unit) is RegexpConcat:
+        chars = [flatten_regex_or_tree(u) for u in unit.units]
+        string = b""
+        for c in chars:
+            if c is None:
+                return None
```
msm-code, line 180: shouldn't it be `continue`, and shouldn't we return more than one string here? Think, something like:

```
asdf(^&*)qwer
```
(where `^&*` represents anything we don't currently support)

the result should be `"asdf" & "qwer"` instead of None IMO

More practical example:
```python
strings: $a = /(tests? code)/
```

src/lib/yaraparse.py
```diff
@@ -153,20 +159,114 @@ def ursify_nocase_bytes(raw: bytes) -> UrsaExpression:
...
+                        s,
+                        is_ascii=is_ascii,
+                        is_wide=is_wide,
+                        is_nocase=is_nocase,
+                    )
+                    for s in or_strings
+                ]
+                strings.append(UrsaExpression.or_(*or_ursa_strings))
+
+        if joined_string and (
```
msm-code, line 213: why the `joined_string`? As far as I can see, it's only non-empty when `if type(unit) is RegexpText`, and then it's immediately changed back to `b""`. Maybe this if can be moved inside the `if type(unit) is RegexpText` branch?

src/lib/yaraparse.py
```diff
@@ -153,20 +159,114 @@ def ursify_nocase_bytes(raw: bytes) -> UrsaExpression:
...
+                or_ursa_strings = [
+                    ursify_plain_string(
+                        s,
+                        is_ascii=is_ascii,
+                        is_wide=is_wide,
+                        is_nocase=is_nocase,
+                    )
+                    for s in or_strings
+                ]
+                strings.append(UrsaExpression.or_(*or_ursa_strings))
```
msm-code, line 211: One popular thing in real-world regexes is apparently `/(T)his/`.

Quik grep:

```
malware/RAT_Gh0st.yar:        $ = /(W)inSta0\\Default/
malware/RAT_Gh0st.yar:        $ = /(#)32770/
malware/RAT_Gh0st.yar:        $ = /(#)32771/
malware/RAT_Gh0st.yar:        $ = /(#)32772/
malware/RAT_Gh0st.yar:        $ = /(#)32774/
malware/RAT_Njrat.yar:        $string1 = /(F)romBase64String/
malware/RAT_Njrat.yar:        $string2 = /(B)ase64String/
malware/RAT_Njrat.yar:        $string3 = /(C)onnected/ wide ascii
malware/RAT_Njrat.yar:        $string4 = /(R)eceive/
```
and lots more.

Is it possible and reasonable to special-case groups with only a single item? Or is it too hard or ugly to implement.

I've looked quickly and it looks like the latter (it's hard to tell when this optimisation is possible), but just asking.

src/tests/yararules/testdata/regular_expressions.txt
```diff
@@ -1 +1 @@
-(({}))
+(((({6d64353a20}) & (({6f6e} | {6f6666}) & {73746174653a20}))))
```
msm-code, line 1: Since we have only one rule with regex, can you add one or two more tests? To make sure we don't introduce regressions in the future. The easiest way is probably

```
git clone https://github.com/Yara-Rules/rules/
cp $goodrules ./src/tests/yararules/testdata
python3 ./src/tests/yararules/generate_yaraparse_result_files.py $goodrules
```

nazywam: done

src/lib/yaraparse.py
nazywam: yeah, would be kinda ugly and hacky in current model

src/lib/yaraparse.py
nazywam: it's not immediately changed, it's used to collect the single characters and separate them when another type of node occurs

PR #175, opened 2020-05-15
Show query time after the query is done

msm-code:
Right now it's not possible to check how long the query took,
and it's a pretty interesting information (at least for me, the developer).

- [yes they are and no i didn't] I've added automated tests for my change (if applicable, optional)
- [-] I've updated documentation to reflect my change (if applicable)

Information when the query was finished is not saved and shown anywhere

That small "took 19s" in the lower right corner under the progress bar

ss

**Test plan**
Run a query, ensure that `took XXXs` text is shown after the query is done (also when it's cancelled or it errors out).

### DIFF
--- a/src/mqueryfront/src/QueryTimer.js
+++ b/src/mqueryfront/src/QueryTimer.js
@@ -37,8 +37,13 @@ class QueryTimer extends Component {
     }
 
     render() {
-        if (!this.props.job.submitted || this.props.isFinished) {
+        if (!this.props.job.submitted) {
             return null;
+        }
+
+        if (this.props.isFinished) {
+            const duration = this.props.job.finished - this.props.job.submitted;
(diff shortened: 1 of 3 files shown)


src/mqueryfront/src/QueryTimer.js
```diff
@@ -37,10 +37,15 @@ class QueryTimer extends Component {
...
 
     render() {
-        if (!this.props.job.submitted || this.props.isFinished) {
+        if (!this.props.job.submitted) {
             return null;
         }
 
+        if (this.props.isFinished) {
+            const duration = this.props.job.finished - this.props.job.submitted;
+            return <i>took {this.getRenderTime(duration)}</i>;
```
ITAYC0HEN, line 46:
```suggestion
            return <i>Duration: {this.getRenderTime(duration)}</i>;
```

PR #176, opened 2020-05-16
Use dropdown menu for download-all

ITAYC0HEN:

The current behavior is that two icons are presented in the Results table. In order to figure what they are, the user needs to hover them and wait for a tooltip to show. Then, a click will download zipped files oe a text file.

image

The new behavior is a little bit cleaner in the UI, save place and using a more standard icon for downloading. Upon clicking, a dropdown menu will be opened, shows two options for the user.

image

Peek 2020-05-16 20-46

image

**Test plan**
1. Check the feature in different browsers. All modern browsers should support it
2. Check that the zip file has the correct name
3. Check that the zip file is downloaded correctly and that all the results are there
4. Check that the txt file has the correct name
5. Check that the txt file is downloaded correctly and that all the hash are presented there
6. Test the feature on a mobile phone
7. Make sure that this new PR wasn't the one to cause the "103%" value in the progress bar on my screenshots because I promise you it wasn't me I swear that this has nothing to do with my PR and that all my screenshots have "off by X%" because of unrelated bugs and that my mquery instance is always trying the best to find more and more samples and even tries to escape the /mnt/samples and be free to search on the internet for files matching my query because why not


msm-code:
> Check the feature in different browsers. All modern browsers should support it

:+1: 

> Check that the zip file has the correct name
Check that the zip file is downloaded correctly and that all the results are there
Check that the txt file has the correct name
Check that the txt file is downloaded correctly and that all the hash are presented there

No changes here

> Test the feature on a mobile phone

It's one of the few things that worked flawlessly on (my) mobile phone

> Make sure that this new PR wasn't the one to cause the "103%" value in the progress bar on my screenshots because I promise you it wasn't me I swear that this has nothing to do with my PR and that all my screenshots have "off by X%" because of unrelated bugs and that my mquery instance is always trying the best to find more and more samples and even tries to escape the /mnt/samples and be free to search on the internet for files matching my query because why not

:fearful:

PR #177, opened 2020-05-17
Fix race condition for queries on the last dataset

msm-code:

Query can get stuck on the last dataset (so not all files are processed)

The not-so-sneaky race condition was here:

```
dataset = self.db.get_next_search_dataset(self.group_id, job_id)
if dataset is None:
	logging.info("Nothing to query, returning...")
	return

...

self.db.update_job_files(job_id, file_count)
```

`get_next_search_dataset` decrements the number of datasets left
(potentially to zero), and increments number of files to be processed
later. So if ursadb is slower than running yara on the results,
there is a time window where the job appears to be done (all files
processed, no datasets left ot query) when in fact the last dataset
is still being queried.

**Test plan**
Run a complicated query with a good recall. Reproducing it is hard on my local machine, though it's 100% reproducible on my server with a lot of files.

### DIFF
--- a/src/daemon.py
+++ b/src/daemon.py
@@ -105,6 +105,7 @@ class Agent:
         self.db.update_job_files(job_id, file_count)
         self.db.agent_start_job(self.group_id, job_id, iterator)
         self.db.agent_continue_search(self.group_id, job_id)
+        self.db.dataset_query_done(job_id)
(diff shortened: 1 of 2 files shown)


msm-code:
Depends on #178

PR #178, opened 2020-05-17
Count datasets left to query

msm-code:
Sometimes queries take a long time - this makes progressbar
a bit useless. Try to explain why to the user with this small change.

Because of the fact that yara results are streamed now, when the number of yara matches is low it's possible to see things like `processing 17/17`.

There is a small badge next to the counter - numbre of datasets left to query:

ss

With tooltip, the best thing since sliced bread:

ss

**Other things that I've considered**

Instead of having this badge, show this on the progress bar with some advanced (fourth grade) math. For example if only half of the (files in) datasets were queried, scale the progressbar to half width and fill the rest with `gray` colour.  

**Test plan**

Run a query and ensure that it still works.
Ensure that the number of datasets to query only decreases, and disappears when it reaches zero

### DIFF
--- a/src/db.py
+++ b/src/db.py
@@ -101,6 +101,7 @@ class Database:
             files_errored=int(data.get("files_errored", 0)),
             iterator=data.get("iterator", None),
             taint=data.get("taint", None),
+            datasets_to_query=data.get("datasets_to_query", 0),
         )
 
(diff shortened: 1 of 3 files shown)


msm-code:
@ITAYC0HEN I know you're a big fan of my UI design, maybe you want to weigh on this

PR #182, opened 2020-05-17
Disable timeouts during indexing

msm-code:
Indexing can take a long time, and sometimes workers can all
be busy for a while. We don't want to crash whole indexing because one
of the indexing jobs took more than 1 second to query backend for a list of
datasets.

- [n/a] I've tested my changes by building and running mquery, and testing changed functionality (if applicable)

One of the queries during indexing can timeout, which will bring down the whole indexing operation.

No timeouts for `topology` requests.

**Test plan**
You probably have to just believe me. Or maybe run utils.index script with more workers than configured in ursadb instance - one of them will timeout and crash the job.

### DIFF
--- a/src/utils/index.py
+++ b/src/utils/index.py
@@ -56,7 +56,9 @@ def index_files(
     ursa_url, types, tags, batch, compact_threshold = proc_params
     ursa = UrsaDb(ursa_url)
 
-    current_datasets = len(ursa.topology()["result"]["datasets"])
+    current_datasets = len(
+        ursa.execute_command("topology;")["result"]["datasets"]
+    )
     if current_datasets > compact_threshold:
         ursa.execute_command("compact smart;")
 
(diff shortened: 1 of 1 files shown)


src/utils/index.py
```diff
@@ -56,7 +56,9 @@ def index_files(
     ursa_url, types, tags, batch, compact_threshold = proc_params
     ursa = UrsaDb(ursa_url)
 
-    current_datasets = len(ursa.topology()["result"]["datasets"])
+    current_datasets = len(
+        ursa.execute_command("topology;")["result"]["datasets"]
```
psrok1, line 60: `UrsaDb.topology`, `UrsaDb.status` and `UrsaDb.execute_command` have almost the same code (the only difference is turned off timeout in the last one). Maybe should be refactored in the future?

msm-code: :weltschmerz2: yeah, good idea

PR #185, opened 2020-05-18
feature/#158 and several warnings removed

### LINKED ISSUE #158: Clicking on the logo should take you to a clean page
**Environment information**
* Mquery version (from the /status page):
* Ursadb version (from the /status page):
* Installation method:
  - [x] Generic docker-compose
  - [x] Dev docker-compose
  - [ ] Native (from source)
  - [ ] Other (please explain)

**Reproduction Steps + Actual behaviour**
When you are is a specific task\query and clicking on the Cert.pl logo on the top of the page, t
it does nothing except changing the url.

**Expected behaviour**

I expect the page to go for a clean query page so I can write a new query

**Screenshots**

The following gif shows that clicking the logo doesnt work as expected. and when opening in new tab it does
Peek 2020-05-11 18-53

JaroPowerH:

fixes several warnings.

Issue158


src/mqueryfront/src/QueryPage.js
```diff
@@ -55,6 +56,30 @@ class QueryPage extends Component {
...
+        if (
+            this.props.match.path === "/" &&
+            this.props.location.state &&
+            this.props.location.state.fromNavigation &&
+            (typeof prevProps.location.state === "undefined" ||
+                typeof prevProps.location.state.fromNavigation === "undefined")
+        ) {
+            //reset page to unitial state
+            this.setState({
+                mode: "query",
```
msm-code, line 69: this is the same state as in the constructor. maybe move this to some variable like `DEFAULT_STATE`?

src/mqueryfront/src/QueryPage.js
```diff
@@ -79,8 +104,8 @@ class QueryPage extends Component {
             queryError: null,
             queryPlan: null,
             qhash: newQhash,
-            matches: [],
-            job: [],
+            matches: null,
```
msm-code, line 107: Why should matches be null instead of `[]`? Matches represent a `list of matches`, so empty list sounds like a good default. But if it works correctly (or if there's something I don't see) i think it's OK too.

src/mqueryfront/src/QueryMonaco.js
```diff
@@ -74,6 +74,9 @@ class QueryMonaco extends Component {
         if (this.props.error !== prevProps.error) {
             this.setError(...this.props.error);
         }
+        if (prevProps.rawYara !== "" && this.props.rawYara === "") {
```
ITAYC0HEN, line 77: what triggered this change? was something broken?

src/mqueryfront/src/QueryPage.js
JaroPowerH: Good idea!. I'll use INITIAL_STATE constant.

src/mqueryfront/src/QueryPage.js
JaroPowerH: An empty list is a good default for the match list. But the code used two defaults: [] and null. I unified the code to consistently use null everywhere.

src/mqueryfront/src/QueryMonaco.js
JaroPowerH: It should be enough to set attribue `value=""` to `<Editor /> `component to clear its content. But it didn't work. The `value` attribute is set to `{this.props.rawYara}`. That's why I added code to clear editor content when corresponding prop.rawYara changes from non empty string to empty string.

src/mqueryfront/src/QueryPage.js
JaroPowerH: By the way, it wasn't exactly the same piece of code. In constructor the `qhash` variable was used for `qhash` attribute.

src/mqueryfront/src/Navigation.js
```diff
@@ -10,9 +10,13 @@ class Navigation extends Component {
     }
 
     render() {
+        const toHome = {
```
psrok1, line 13: Forced remount is not the correct way to fix that issue. Component can't reload correctly because there is no correct state transition from "job" mode to "query" mode.

QueryPage code has few issues related with that bug:
- state is derived from `this.props.match.params.hash` in incorrect way (this props should be used in all places where `qhash` is needed instead of putting it into initial state and relying only on state value. That's why we can't listen for path changes)
- "mode" should be probably deduced from "this.props.match.params.hash". It seems that view goes into the "job mode" only when route `query/:hash` is active.

QueryPage should be refactored to provide correct reaction for path change.

src/mqueryfront/src/QueryPage.js
JaroPowerH: INITIAL_STATE usage in contructor will be:
```
        let initialState = { ...INITIAL_STATE };
        if (this.props.match.params.hash) {
            initialState.qhash = this.props.match.params.hash;
        }
        this.state = initialState;

```

src/mqueryfront/src/QueryMonaco.js
JaroPowerH: This code is unnecessary if the Monaco component is successfully initialized. I will get rid of this code.

src/mqueryfront/src/Navigation.js
JaroPowerH: This code just pass additional info to `QueryPage` (to  component and does not force remount.
With this additional information, the component can distinguish whether the path change has occurred from outside or from within.

psrok1: Follow-up from conversation: Ok, this is still useful when we want to clean-up Yara query form after clicking logo or 'Query' navitem

msm-code:
Doesn't fix the main issue for me completely (#158), but @psrok1 said he'll work on this basing on this code, so i think it's ok to merge this.

PR #199, opened 2020-05-20
Refactor of QueryPage and related components

psrok1:

- Using props for getting qhash from path (`/query/<hash>`) instead of derived state
- Removed redundant `this.state.mode`: QueryPage is considered to be in "job" mode always if `queryHash` is set.
- Query submission method has been lifted from `QueryField` component to the `QueryPage` component (`handleQuery` => `this.props.submitQuery`)
- Instead of setting `internal` location state for all internal redirects, `{ editQuery: this.queryHash }` is set when we want to go to `/` and keep rawYara from the current job (`QueryPage.editQuery` method).

Other (minor) changes:
- Methods that send requests to API are now `async`
- Removed unused methods `QueryField.handleInputChange` and `QueryField.componentDidMount` and unnecessary `QueryField.handleYaraChanged`

**Fixed the following bugs:**
- If logo or `Edit` was clicked during the job - timeout was still active, because it was cleared only when component unmounts
- `datasets` were wiped from state after clicking logo or `Query` navitem (they're fetched only after mount)

### DIFF
--- a/src/mqueryfront/src/QueryField.js
+++ b/src/mqueryfront/src/QueryField.js
@@ -1,93 +1,12 @@
 import React, { Component } from "react";
-import axios from "axios";
-import { API_URL } from "./config";
 import QueryMonaco from "./QueryMonaco";
 
 class QueryField extends Component {
-    constructor(props) {
-        super(props);
-
-        this.state = {
-            selectedTaint: null,
-            error: null,
-        };
-
-        this.handleInputChange = this.handleInputChange.bind(this);
-        this.handleYaraChanged = this.handleYaraChanged.bind(this);
-        this.handleQuery = this.handleQuery.bind(this);
-        this.handleEdit = this.handleEdit.bind(this);
-    }
-
-    componentDidMount() {}
-
-    selectTaint(newTaint) {
-        this.setState({
-            selectedTaint: newTaint,
-        });
-    }
-
     describeTaint() {
-        if (this.state.selectedTaint == null) {
+        if (this.props.selectedTaint == null) {
             return "everywhere";
         }
-        return this.state.selectedTaint;
-    }
-
-    handleYaraChanged(value) {
-        this.props.updateYara(value);
-    }
-
-    handleQuery(event, method, priority) {
-        const yara = this.props.rawYara;
-        axios
-            .create()
-            .post(API_URL + "/query", {
-                raw_yara: yara,
-                method: method,
-                priority: priority,
-                taint: this.state.selectedTaint,
-            })
-            .then((response) => {
-                if (method === "query") {
-                    this.props.updateQhash(response.data.query_hash, yara);
-                } else if (method === "parse") {
-                    this.props.updateQueryPlan(response.data, yara);
-                }
-            })
-            .catch((error) => {
-                let err = error.toString();
-
-                if (error.response) {
-                    err = error.response.data.detail;
-                    // Dirty hack to parse error lines from the error message
-                    // Error format: "Error at 4.2-7:" or  "Error at 5.1:"
-                    let parsedError = err.match(
-                        /Error at (\d+).(\d+)-?(\d+)?: (.*)/
-                    );
-                    if (parsedError) {
-                        this.setState({ error: parsedError });
-                    }
-                }
-
-                this.props.updateQueryError(err, this.props.rawYara);
-            });
-
-        event.preventDefault();
-    }
-
-    handleInputChange(event) {
-        const target = event.target;
-        const value =
-            target.type === "checkbox" ? target.checked : target.value;
-        const name = target.name;
-
-        this.setState({
-            [name]: value,
-        });
-    }
-
-    handleEdit(event) {
-        this.props.updateQhash(null);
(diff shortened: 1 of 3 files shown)


src/mqueryfront/src/QueryField.js
```diff
@@ -1,93 +1,12 @@
...
-
-    selectTaint(newTaint) {
-        this.setState({
-            selectedTaint: newTaint,
-        });
-    }
-
     describeTaint() {
-        if (this.state.selectedTaint == null) {
+        if (this.props.selectedTaint == null) {
```
msm-code, line 6: maybe `this.props.selectedTaint or "everywhere"` (I know it was originally written by me, but sounds like an improvement - thoughts?)

src/mqueryfront/src/QueryField.js
```diff
@@ -97,9 +16,10 @@ class QueryField extends Component {
                     <button
                         type="button"
                         className="btn btn-success btn-sm"
-                        onClick={(event) =>
-                            this.handleQuery(event, "query", "medium")
-                        }
+                        onClick={(event) => {
+                            event.preventDefault();
```
msm-code, line 20: Why are there `preventDefault`s everywhere? I think this button is not in any form, so it shouldn't change its behaviour?

src/mqueryfront/src/QueryPage.js
```diff
@@ -77,127 +98,115 @@ class QueryPage extends Component {
...
+                "/matches/" +
+                this.queryHash +
+                "?offset=" +
+                OFFSET +
+                "&limit=" +
+                LIMIT
+        );
+        if (response) {
+            const { job, matches } = response.data;
+            return { job, matches };
```
msm-code, line 146: is it different than `return response.data`?

psrok1: 🤔 Yup, this looks like unnecessary filter for these two keys..

src/mqueryfront/src/QueryField.js
psrok1: Agree, this is unnecessary for buttons.

src/mqueryfront/src/QueryField.js
psrok1: Sure, looks like good case for short circuiting!

src/mqueryfront/src/QueryPage.js
```diff
@@ -24,49 +23,71 @@ class QueryPage extends Component {
...
-            this.props.location.key !== prevProps.location.key &&
-            (typeof this.props.location.state === "undefined" ||
-                typeof this.props.location.state.internal === "undefined" ||
-                !this.props.location.state.internal)
-        ) {
-            this.setState(INITIAL_STATE);
+        const prevQueryHash = prevProps.match.params.hash;
+        if (this.queryHash) {
+            if (prevQueryHash !== this.queryHash) {
+                // Went to the job mode or switched to another job
```
ITAYC0HEN, line 76: what is "job mode"?

ITAYC0HEN: Is it the shift from writing a query to executing a query?

src/mqueryfront/src/QueryPage.js
```diff
@@ -77,127 +98,111 @@ class QueryPage extends Component {
...
+        if (this.state.queryError) {
+            // Dirty hack to parse error lines from the error message
+            // Error format: "Error at 4.2-7:" or  "Error at 5.1:"
+            let parsedError = this.state.queryError.match(
+                /Error at (\d+).(\d+)-?(\d+)?: (.*)/
+            );
+            if (parsedError) return parsedError;
+        }
+        return [];
     }
```
ITAYC0HEN, line 189: is this right to use getter here? the object is manipulated and it doesn't seem like it is exclusively bound to "queryError" which is used on other places. But hey! I am not a JS coder so I dont know more than you do

src/mqueryfront/src/QueryPage.js
psrok1: Yes. Previously there was `this.state.mode` element that was in two states depending on the phase: `query` (editing) or `job` (executing). Now, it's deduced from the path (`/` for `query` and `/query/<hash>` for `job`). I decided that I'd still refer to that `mode` state in comments to improve code readability.

ITAYC0HEN: looks good, consider improving the comment maybe. 
nothing too important, i just wondered

src/mqueryfront/src/QueryPage.js
psrok1: Actually I was also in doubt how this `parsedError` should be implemented. I've just lifted it from QueryField along with query submission method and placed it in getter, because its value seemed to be directly derived from queryError message. But maybe this could be placed in QueryMonaco because that regex has strong dependency with `setError` arguments. Such design decisions are always pain in the neck :/

ITAYC0HEN: there is no right answer, flip a coin or let `random.choose` to decide :P

PR #202, opened 2020-05-27
Fix/keep chosen taint options

### LINKED ISSUE #139: Multiselect for sample tags

dskwhitehat:

Taint selection was cleared after running query.

Taint selection is not cleared after running query.

**Test plan**
1. Add taints to datasets and select it with multiselect.
2. Try to query with some rule
Expected result:
You should see that taints selection is not cleared.
3. Click on CERT.PL
Expected result:
Taints selection should be cleared.

**Closing issues**

fixes #139

### DIFF
--- a/src/mqueryfront/src/QueryField.js
+++ b/src/mqueryfront/src/QueryField.js
@@ -26,7 +26,7 @@ class QueryField extends Component {
                 <ReactMultiSelectCheckboxes
                     onChange={this.props.handleChange}
                     options={options}
-                    defaultValue={this.props.selectedTaints}
+                    value={this.props.selectedTaints}
                     placeholderButtonLabel={placeHolder}
                 />
             );
--- a/src/mqueryfront/src/QueryField.js
+++ b/src/mqueryfront/src/QueryField.js
@@ -4,21 +4,18 @@ import ReactMultiSelectCheckboxes from "react-multiselect-checkboxes";
 
 class QueryField extends Component {
     render() {
-        var options = this.props.availableTaints.map(function (obj) {
-            return {
-                label: obj,
-                value: obj,
-            };
-        });
+        const options = this.props.availableTaints.map((obj) => ({
+            label: obj,
+            value: obj,
+        }));
 
-        var multiselect = null;
-        var placeHolder = "everywhere";
-
+        let multiselect = null;
+        let placeholder = "everywhere";
         if (this.props.selectedTaints.length) {
-            placeHolder = this.props.selectedTaints.map(function (obj) {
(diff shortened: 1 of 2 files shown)


src/mqueryfront/src/QueryField.js
```diff
@@ -12,13 +12,22 @@ class QueryField extends Component {
         });
 
         var multiselect = null;
+        var placeHolder = "everywhere";
```
msm-code, line 15: `placeholder` instead of `placeHolder` (for consistency with other places in code, also it's one word not two)

src/mqueryfront/src/QueryField.js
```diff
@@ -12,13 +12,22 @@ class QueryField extends Component {
...
+
+        if (this.props.selectedTaints.length) {
+            placeHolder = this.props.selectedTaints.map(function (obj) {
```
msm-code, line 18: it's better to use `obj => obj.value` here

src/mqueryfront/src/QueryField.js
```diff
@@ -12,13 +12,22 @@ class QueryField extends Component {
         });
 
         var multiselect = null;
+        var placeHolder = "everywhere";
+
+        if (this.props.selectedTaints.length) {
```
msm-code, line 17: is this check necessary? if the state is always a list (empty or non-empty) it shouldn't be necessary.

src/mqueryfront/src/QueryPage.js
```diff
@@ -88,6 +87,9 @@ class QueryPage extends Component {
                 datasets: this.state.datasets,
                 rawYara: editMode ? this.state.rawYara : "",
             });
+            if (!editMode) {
+                window.location.reload(true);
```
msm-code, line 91: (as discussed on slack, we have to change this. Will take a look later).

src/mqueryfront/src/QueryField.js
BonusPlay, line 15: `var` :angry:

BonusPlay: Don't **ever** use `var`, prefer `let` or `const`.

PR #205, opened 2020-05-28
Add a flag for minimal file size

msm-code:

There is no way to specify minimum file size when indexing

It's possible to specify minimal file size for indexing. This is useful when bucketing indexed files by size when indexing.

**Test plan**
Run the `index.py` script with min-file-size flag, and ensure that it doens't index files smaller than it.

fixes #issuenumber

### DIFF
--- a/src/utils/index.py
+++ b/src/utils/index.py
@@ -39,10 +39,17 @@ def walk_directory(dir: Path, ignores: List[str]) -> Iterator[Path]:
 
 
 def find_new_files(
-    existing: Set[str], files_root: Path, mounted_as: str, max_file_size: int
+    existing: Set[str],
+    files_root: Path,
+    mounted_as: str,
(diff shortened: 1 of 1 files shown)


src/utils/index.py
```diff
@@ -210,6 +220,12 @@ def main() -> None:
         help="Path to the samples to be indexed, as seen by ursadb (if different).",
         default=None,
     )
+    parser.add_argument(
+        "--min-file-size-mb",
+        type=int,
+        help="Minimum file size, in MB, to index. 0 By default.",
+        default=0,
```
psrok1, line 227: Maybe 'indexing empty files' is not the best default? 🤔

msm-code: it's not wrong though (principle of least surprise suggests that we shouldn't ignore any file in the directory)

PR #212, opened 2020-06-04
Job status now changes to done and job is finished when nothing is in…

### LINKED ISSUE #192: Job status never changes to `done` when the database is empty.
**Environment information**
* Mquery version (from the /status page): 1.1.0
* Ursadb version (from the /status page): 1.4.0+7867438
* Installation method:
  - [x] Generic docker-compose
  - [x] Dev docker-compose
  - [ ] Native (from source)
  - [ ] Other (please explain)

**Reproduction Steps**

- Start a new mquery instance without indexing anything
- Run any kind of job

**Expected behaviour**

The job finishes immediately

**Actual behaviour the bug**

The job is stuck in the `processing` state.

**Screenshots**

ss

**Additional context**

This happens, because that corner case is not handled in the `daemon.py` correctly. When there are no datasets we should immediately change job status to success.

dskwhitehat:
…dexed.

Job status is processing after making query when nothing is indexed.

Job status is done after making query with no indexes.

**Test plan**
Try to make reproduction steps like in #192. Make regression test with files indexed. 

**Closing issues**

fixes #192

### DIFF
--- a/src/daemon.py
+++ b/src/daemon.py
@@ -67,12 +67,6 @@ class Agent:
         """
         logging.info("Parsing...")
 
-        db_topology = self.ursa.topology()
-        logging.info("No datasets found. Finish the job and return...")
-        if not db_topology["result"]["datasets"].keys():
-            self.db.agent_finish_job(job_id)
-            return
-
         job = self.db.get_job(job_id)
         if job.status == "cancelled":
             logging.info("Job was cancelled, returning...")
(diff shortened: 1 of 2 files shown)


src/daemon.py
```diff
@@ -67,6 +67,14 @@ def __search_task(self, job_id: JobId) -> None:
         """
         logging.info("Parsing...")
 
+        result = self.ursa.topology()
```
msm-code, line 70: why move `.topology()` here? this is necessary only for new jobs, so this line should stay where it was.

src/daemon.py
```diff
@@ -67,6 +67,14 @@ def __search_task(self, job_id: JobId) -> None:
...
+        logging.info(
+            "If there are no datasets indexed finish job and return..."
```
msm-code, line 72:
```suggestion
            "No datasets found. Finish the job and return..."
```

src/daemon.py
dskwhitehat: I use it to check if any datasets are indexed so this is needed for whole __search_task function. But if you want I can change it but then we will need to do the same operation self.ursa.topology() twice.

PR #219, opened 2020-06-16
Add rule name filter for matched samples

### LINKED ISSUE #78: Filter query results based on rule-name
**Description**
I think it is a good idea to let the user interact a little bit more with the results page. This including secondary filtering.

For example, a query can contain multiple rules, and each sample that showed in the matches table has a label(s) with the name(s) of the matched rule(s).

This can be shown in the following image.
image

I'd like to a more dynamic approach for the matches table, in which the user can click on a label (rule name) and the table will be filtered based on the selected rule.

Not surprisingly, this is implemented in [mwdb](https://mwdb.cert.pl/) in a very nice way. The user can click on a tag and it will immediately filter.

KWMORALE:

Implement rule name filter which allows you to filter samples after clicking on the name of the rule in the results.

**Test plan**
Run different queries at the same time that match different samples and check if filtering works.

**Closing issues**

fixes #78

### DIFF
--- a/src/mqueryfront/src/App.css
+++ b/src/mqueryfront/src/App.css
@@ -83,3 +83,7 @@
     cursor: pointer;
     color: var(--blue) !important;
 }
+
+.cursor-pointer{
+    cursor: pointer;
+  }
--- a/src/mqueryfront/src/App.css
+++ b/src/mqueryfront/src/App.css
@@ -84,6 +84,6 @@
     color: var(--blue) !important;
 }
 
-.cursor-pointer{
+.cursor-pointer {
     cursor: pointer;
-  }
+}
--- a/src/mqueryfront/src/components/FilterIcon.js
+++ b/src/mqueryfront/src/components/FilterIcon.js
@@ -17,7 +17,7 @@ const FilterIcon = (props) => {
 
 FilterIcon.defaultProps = {
     size: "xs",
-    color: "black",
+    color: undefined,
     tooltipMessage: "filter",
 };
 
--- a/src/mqueryfront/src/components/FilterIcon.js
+++ b/src/mqueryfront/src/components/FilterIcon.js
@@ -17,7 +17,6 @@ const FilterIcon = (props) => {
 
 FilterIcon.defaultProps = {
     size: "xs",
-    color: undefined,
     tooltipMessage: "filter",
 };
 
--- a/src/mqueryfront/src/query/QueryMatches.js
+++ b/src/mqueryfront/src/query/QueryMatches.js
@@ -94,7 +94,7 @@ const QueryMatches = (props) => {
                             </span>
                             {filters.length > 0 && (
                                 <span className="border rounded p-1 pull-right text-secondary">
-                                    <FilterIcon color="text-secondary" />
+                                    <FilterIcon />
                                     {filtersHead}
                                 </span>
                             )}
--- a/src/mqueryfront/src/query/QueryMatches.js
+++ b/src/mqueryfront/src/query/QueryMatches.js
@@ -31,7 +31,7 @@ const QueryMatches = (props) => {
             }
         })
         .map((match, index) => {
-            const download_url =
+            const downloadUrl =
                 API_URL +
                 "/download?job_id=" +
                 encodeURIComponent(qhash) +
@@ -44,13 +44,23 @@ const QueryMatches = (props) => {
                 <QueryMatchesItem
                     key={match.file}
                     match={match}
-                    download_url={download_url}
+                    download_url={downloadUrl}
                     filters={filters}
                     setFilter={setFilter}
                     changeFilter={updateFilter}
                 />
             );
         });
+
+    const filtersHead = filters.map((v) => (
+        <span
+            key={v}
+            className="badge badge-pill badge-primary ml-1 mt-1  cursor-pointer"
+            onClick={() => updateFilter(v)}
+        >
+            {v}
+        </span>
+    ));
 
     const downloadDropdownList = [
         {
@@ -82,15 +92,7 @@ const QueryMatches = (props) => {
                                     itemList={downloadDropdownList}
                                 />
                             </span>
-                            {filters.map((v) => (
-                                <span
-                                    key={v}
-                                    className="badge badge-pill badge-primary ml-1 mt-1"
-                                    onClick={() => updateFilter(v)}
-                                >
-                                    {v}
-                                </span>
-                            ))}
+                            {filtersHead}
                         </th>
                     </tr>
                 </thead>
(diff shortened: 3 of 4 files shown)


KWMORALE:
Peek 2020-06-16 18-48

KWMORALE:
I will check it tomorrow on some real files and questions.
Double click prevention still needs to be implemented.

msm-code:
Random note - use `cursor: pointer` for tags to show that they are clickable

src/mqueryfront/src/query/QueryMatches.js
```diff
@@ -9,24 +9,48 @@ import { PT_MATCHES, PT_PAGINATION } from "../queryUtils";
...
+
+    const matchesList = matches
+        .filter((match) => {
+            if (
+                filters.length > 0 &&
+                match.matches.some((v) => filters.includes(v))
+            ) {
+                return match;
+            } else {
+                return match;
```
msm-code, line 30: Both branches return the same thing, so this all is a no-op. Is this `.filter` necessary?

src/mqueryfront/src/query/QueryMatches.js
```diff
@@ -9,24 +9,48 @@ import { PT_MATCHES, PT_PAGINATION } from "../queryUtils";
...
+            }
+        })
+        .map((match, index) => {
+            const download_url =
```
msm-code, line 34: should be `downloadUrl`

src/mqueryfront/src/App.css
```diff
@@ -83,3 +83,7 @@
     cursor: pointer;
     color: var(--blue) !important;
 }
+
+.cursor-pointer{
+    cursor: pointer;
+  }
```
psrok1, line 89:
```suggestion
.cursor-pointer {
    cursor: pointer;
}

```

KWMORALE:
In future we could easily add this functionality to the metadata badges.

ITAYC0HEN:
Looks good!
What about having something that indicates that the labels at the top are filtered?
image


When looking at such image it isn't clear ^

ITAYC0HEN:
Sketch example:
image


```
<span class="border rounded p-1 pull-right">
    <i class="fa fa-filter text-secondary mr-2"></i>
    {filtersHead}
</span>
````

**Comment by ITAYC0HEN** [comment]
Another thing that wasn't obvious to me, but isn't necessarily bad, is that the condition of the filtered rules is OR and not AND

**Comment by KWMORALE** [comment]
Screenshot from 2020-06-19 16-13-31
Looks better

File: src/mqueryfront/src/query/QueryMatches.js [hunk 4]
```diff
@@ -93,8 +93,8 @@ const QueryMatches = (props) => {
                                 />
                             </span>
                             {filters.length > 0 && (
-                                <span class="border rounded p-1 pull-right">
-                                    <FilterIcon color="grey" />
+                                <span class="border rounded p-1 pull-right text-secondary">
+                                    <FilterIcon color="text-secondary" />
```
ITAYC0HEN, line 97: The color isn't necessary if you used `text-secondary` on the span.
```suggestion
                                    <FilterIcon />
```

ITAYC0HEN: (and it wouldn't work not work, because `text-secondary` is a class and not color)

src/mqueryfront/src/query/QueryMatches.js
```diff
@@ -58,6 +92,12 @@ const QueryMatches = (props) => {
                                     itemList={downloadDropdownList}
                                 />
                             </span>
+                            {filters.length > 0 && (
+                                <span class="border rounded p-1 pull-right text-secondary">
```
psrok1, line 96: There's no `class` in React elements, use `className`
```suggestion
                                <span className="border rounded p-1 pull-right text-secondary">
```

src/mqueryfront/src/query/QueryMatchesItem.js
```diff
@@ -22,7 +26,11 @@ const QueryMatchesItem = (props) => {
         ));
 
     const matchBadges = Object.values(matches).map((v) => (
-        <span key={v} className="badge badge-pill badge-primary ml-1 mt-1">
+        <span
+            key={v}
+            className="badge badge-pill badge-primary ml-1 mt-1 cursor-pointer"
+            onClick={() => addFilter(v)}
```
psrok1, line 32: Arrow function calling another arrow function with wrong argument name (suggesting that we're dealing with event when it's a match). It should be simplified.
```suggestion
            onClick={() => props.changeFilter(v)}
```

src/mqueryfront/src/query/QueryMatchesItem.js
```diff
@@ -8,6 +8,10 @@ const QueryMatchesItem = (props) => {
     const { match, download_url } = props;
     const { matches, meta, file } = match;
 
+    const addFilter = (event) => {
+        props.changeFilter(event);
+    };
+
```
psrok1, line 14: Remove it in the same batch as `onClick` suggestion
```suggestion
```

src/mqueryfront/src/query/QueryMatches.js
KWMORALE: Interesting, but the color (text-secondary) for an icon is set only when it is defined in the class and icon argument. I just leave the grey icon color instead.

src/mqueryfront/src/query/QueryMatches.js
```diff
@@ -58,6 +92,12 @@ const QueryMatches = (props) => {
                                     itemList={downloadDropdownList}
                                 />
                             </span>
+                            {filters.length > 0 && (
+                                <span className="border rounded p-1 pull-right">
+                                    <FilterIcon color="grey" />
+                                    {filtersHead}
+                                </span>
```
ITAYC0HEN, line 99: @psrok1 @KWMORALE @JaroPowerH why are FilterIcon and ActionRemove are forced for Black color? This way, setting text-color on span and parent objects will not apply to the icons?

For example, I'd prefer removing `color="grey"` from line 97 in favor of adding `text-secondary` class in line 96. This is currently not possible because FilterIcon's default props set the icon color to black, inlined (see FilterIcon.js), and thus ignoring the color scheme specified by its parent

src/mqueryfront/src/components/FilterIcon.js
```diff
@@ -17,7 +17,7 @@ const FilterIcon = (props) => {
 
 FilterIcon.defaultProps = {
     size: "xs",
-    color: "black",
+    color: undefined,
```
psrok1, line 20: I think we can just delete that defaultProp. Undefined is the same as missing key.
```suggestion
```

KWMORALE:
Screenshot from 2020-06-25 14-15-18
Screenshot from 2020-06-25 14-15-28
looks the same with less code

PR #225, opened 2020-12-11
Remove deprecated time.clock from ursadb client

nazywam:

UrsaDB client  within mquery uses `time.clock` method that was removed in python 3.8. AFAIK this completely breaks the code on python3.8+

`time.clock` is replaced with `time.perf_counter` (added in python 3.3) which should work for all 3.6+ python versions.

**Test plan**
As far as I can see the `"time"` component from `UrsaDb.query` isn't really used anywhere :thinking:
E2E CI tests should make sure that there aren't any runtime errors.

**Closing issues**

### DIFF
--- a/src/lib/ursadb.py
+++ b/src/lib/ursadb.py
@@ -48,7 +48,7 @@ class UrsaDb:
     ) -> Json:
         socket = self.make_socket(recv_timeout=-1)
 
-        start = time.clock()
+        start = time.perf_counter()
         command = "select "
         if taints:
             taints_str = '", "'.join(taints)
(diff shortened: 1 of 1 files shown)


src/lib/ursadb.py
```diff
@@ -76,7 +76,7 @@ def query(
         file_count = res["result"]["file_count"]
 
         return {
-            "time": (end - start) * 1000,
+            "time": (end - start),
```
msm-code, line 79: The resolution is the same, so this should probably still be in ms:

```suggestion
            "time": (end - start) * 1000,
```

msm-code (changes requested):

msm-code (approved):

PR #229, opened 2021-04-26
Fix regex parsing

bzeba:

Regex [escape sequences](https://yara.readthedocs.io/en/stable/writingrules.html#regular-expressions) and character classes aren't parsed correctly. They get parsed as literal strings which results in false negatives, i.e. files which should match don't. 
Simple regex such as `/abc|def/` causes AttributeError while parsing.

Escape sequences are transformed into bytes which they represent, including hex escapes (`\xdd`). Zero-width assertions (`\b` and `\B`) are simply removed from the output. Strings get split around character class tokens (e.g. `\w`) and tokens are removed.

Parsing simple "or" regex doesn't cause an exception anymore. 

**Test plan**
Since it only affects parsing, I think included automated test suffice. Testing the whole system on real examples using the affected features of regular expressions would be useful.


nazywam:
Can confirm that the parser fails on simple OR regexes:
```
Traceback (most recent call last):
  File "src/lib/yaraparse.py", line 516, in <module>
    main()
  File "src/lib/yaraparse.py", line 512, in main
    print(rule.name, rule.parse().query)
  File "src/lib/yaraparse.py", line 99, in parse
    self.__parsed = self.__parse_internal()
  File "src/lib/yaraparse.py", line 92, in __parse_internal
    result = parser.traverse(self.rule.condition)
  File "src/lib/yaraparse.py", line 459, in traverse
    return self.CONDITION_HANDLERS[type(condition)](self, condition)
  File "src/lib/yaraparse.py", line 322, in and_expr
    right = self.traverse(condition.right_operand)
  File "src/lib/yaraparse.py", line 459, in traverse
    return self.CONDITION_HANDLERS[type(condition)](self, condition)
  File "src/lib/yaraparse.py", line 348, in str_expr
    return ursify_string(self.strings[condition.id])
  File "src/lib/yaraparse.py", line 308, in ursify_string
    return ursify_regex_string(string)
  File "src/lib/yaraparse.py", line 236, in ursify_regex_string
    string.unit.units, is_ascii=True, is_nocase=string.is_nocase
AttributeError: 'yaramod.RegexpOr' object has no attribute 'units'
```
(Works with the patch though)

src/lib/yaraparse.py
```diff
@@ -186,44 +235,63 @@ def flatten_regex_or_tree(unit: Any) -> Optional[List[bytes]]:
...
+    unit: Any, is_ascii: bool, is_wide: bool, is_nocase: bool,
+) -> Optional[UrsaExpression]:
+    or_strings = flatten_regex_or_tree(unit)
+    if or_strings and all(s is not None for s in or_strings):
+        or_ursa_strings: List[UrsaExpression] = []
+        for s in or_strings:
+            ursified_s = ursify_regex_text(
+                s, is_ascii=is_ascii, is_wide=is_wide, is_nocase=is_nocase
+            )
+            if ursified_s:
```
nazywam, line 248: I'm wondering if the `else` statement here shouldn't lead to early return as well.
My logic would be that:
* for `and` expressions, we can drop non-parsable strings as this will only lead to possible false positives
* for `or` expressions, we either have to correctly parse all strings or none at all, as dropping even a single one can lead to a false negative

src/lib/yaraparse.py
```diff
@@ -186,44 +235,63 @@ def flatten_regex_or_tree(unit: Any) -> Optional[List[bytes]]:
...
-                    for s in or_strings
-                ]
-                strings.append(UrsaExpression.or_(*or_ursa_strings))
-
-        if joined_string and (
-            type(unit) is not RegexpText or i == len(units) - 1
-        ):
-            strings.append(
-                ursify_plain_string(
+    if type(regex) is not RegexpConcat:
```
nazywam, line 266: What do you think about reversing the if statement here? i.e. converting the other branch to `if type(regex) is RegexpConcat` and then this one to `else`

src/lib/yaraparse.py
```diff
@@ -186,44 +235,63 @@ def flatten_regex_or_tree(unit: Any) -> Optional[List[bytes]]:
         return None
 
 
+def urisfy_regex_or_tree(
+    unit: Any, is_ascii: bool, is_wide: bool, is_nocase: bool,
+) -> Optional[UrsaExpression]:
+    or_strings = flatten_regex_or_tree(unit)
+    if or_strings and all(s is not None for s in or_strings):
```
nazywam, line 242: Nitpicking at this point, but could this be converted as:
```suggestion
    if or_strings and all(or_strings):
```
or do you think this can cause problems with empty/non-truthy strings?

nazywam (commented):
Left some early comments, I'll let @msm-code also have a gander at this.

Thanks for spending your time to work on this!

src/lib/yaraparse.py
bzeba: Great point. Fixed and I've also added a test for this case. I hope you don't mind me just adding more rules to existing test files. Let me know if you want it separated.

src/lib/yaraparse.py
bzeba: I've tested the feasible cases and it shouldn't be a problem. I've inherited this line from the code before, so hopefully it's not an oversight of some kind.

msm-code:
Sorry, I didn't get a notification about this PR. I'll review it tomorrow,

bzeba:
Any updates? :)

Not trying to be annoying or anything, just a reminder in case it just got forgotten.

nazywam (approved):
Sorry this took **so much** time, I think it works great!

The only issue I came across was that regexes with optional hex-encoded characters get parsed incorrectly `/12\x33?4/` -> `(min 1 of (({31325c7833} & {34})))`. But this is actually caused by yaramod and will have to be fixed independently of this PR.

Thanks!

PR #244, opened 2021-12-05
Added utility script for cleaning unneccesary dataset files

alex-ilgayev:
- [X] I've updated documentation to reflect my change (if applicable)

Running the script will iterate over the index directory, and remove files matching to one of dataset formats with unidentified ID. It will skip unmatched files which aren't related to ursadb.
The usage is explained at docs/utils.

**Test plan**
Add a file matching to ursadb dataset format, like `set.bfd8cc40.db.ursa` with ID which isn't preset at `db.ursa`. Run the script, check if that file is being deleted.

### DIFF
--- a/docs/utils/remove_datasets.md
+++ b/docs/utils/remove_datasets.md
@@ -1,0 +1,23 @@
+# remove_datasets.py
+
+Remove unneccesary dataset files in the index directory.
+It doing that by checking the IDs of all files in the directory,
+and validating if they exist in "db.ursa" file.
+
+## Usage
+Update the "index_basepath" variable:
+`index_basepath = "/mnt/index/"`
+
+Run the script without arguments:
+
+```
+$ python3 utils/remove_datasets.py
+```
+
+## Example
+
+```
+$ python3 utils/remove_datasets.py
+INFO:root:Removing hash4.set.bfd8cc44.db.ursa
+...
+```
--- a/src/utils/remove_datasets.py
+++ b/src/utils/remove_datasets.py
@@ -41,7 +41,7 @@ def get_datasets() -> List[str]:
         db = json.load(f)
     datasets = db["datasets"]
     return list(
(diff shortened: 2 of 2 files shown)


williballenthin:
I think this *might* overlap with the functionality provided by the `ursadb_trim` program here: https://github.com/CERT-Polska/ursadb/blob/master/src/Trim.cpp

msm-code:
Thanks for contributing!

Like @williballenthin pointed out, I think this is very similar to `ursadb_trim` command.

FWIW ursadb_trim has some nice features:
 - Dry run
 - Doesn't hardcode `index_basepath = "/mnt/index/"` (this will only work for docker, not for bare-metal deployments)
 - Has no python dependency (so works on ursadb container too)
 - Traverses the db JSON instead of relying on name formats (database doesn't parse the filename - so it's unusual, but legal to rename files on disk to something different)
 - It already exists, so one less thing to maintain :).

It's certainly underdocumented though (it's not even mentioned in the docs :/)

Please take a look at it and tell me if it works for you. If yes, maybe the solution is to improve the documentation. And if there's something missing, maybe I can improve ursadb_trim instead?

alex-ilgayev:
Hey, thanks for the quick response!

Unfortunately, I couldn't run it due to `terminate called after throwing an instance of 'nlohmann::detail::parse_error'
  what():  [json.exception.parse_error.101] parse error at 1: syntax error - unexpected end of input; expected '[', '{', or a literal` error. Reading a bit [here](https://github.com/nlohmann/json/issues/2066) didn't help to understand or resolve it.
  Nevertheless, going over the code, it is doing exactly what I expected from such a tool, so I think we can close this PR.

msm-code:
@alex-ilgayev can you give me example of how your database file looks? Maybe it's a bug in the `ursadb_trim`.

This *sounds* like the parser is trying to parse something that isn't JSON (it opened an empty file). Maybe a file permission issue?

Anyway the error message could be better. I'll work on that.

PR #246, opened 2021-12-07
Update all the requirements

### LINKED ISSUE #100: Add a support for xor ranges, when they are introduced to yaramod
**Feature Category**

- [x] Correctness
- [ ] Performance
- [ ] Other (please explain)

**Describe the problem**

We don't support yara xor modifiers (only xor with range 0-255). This is currently not exposed by upstream.

Fix this in upstream or ask them to do this, and later use this in our code.

### LINKED ISSUE #231: Fix regex parsing (in avast/yaramod)
Discovered during testing a PR.

There is a bug in upstream regex parsing. Already reported here: https://github.com/avast/yaramod/issues/186. We should help yaramod fix this issue (or wait for when they have time).

### LINKED ISSUE #235: Bump dependencies to the current version
**Feature Category**

- [ ] Correctness
- [ ] Performance
- [x] Other (please explain) - Maintenance

**Describe the problem**

We pin all of our dependencies.
Because of that, some of our dependencies are outdated

Bump all the dependencies to the newest possible version. Test the system after that.

"If it's not broken, don't fix it" - maybe, but there are probably hiden bugs that we didn't notice yet. Especially in the UI code.

msm-code:

Dependencies are updated

Suprisingly, everything seems to work after updating the dependenceis (and minor fixes). I'll do more testing later, but I think tihs is ready for review.

**Test plan**
Click around

**Closing issues**

fixes #231 (tested)
fixes #235
fixes #100


src/app.py
```diff
@@ -63,7 +62,7 @@ def download(job_id: str, ordinal: int, file_path: str) -> FileResponse:
     arbitrary files (for example "/etc/passwd").
     """
     if not db.job_contains(JobId(job_id), ordinal, file_path):
-        raise NotFound("No such file in result set.")
+        raise Response("No such file in result set.", status_code=404)
```
nazywam, line 65:
```suggestion
        return Response("No such file in result set.", status_code=404)
```
probably?

nazywam (changes requested):
Looks good, the other failed tests seem unrelated.

What's the usecase of `requirements.plain.txt`, should we maybe put it in a comment somewhere?

msm-code:
Uh oh, thanks for catching this. I've actually fixed this and forgot to commit :facepalm:.

I'll apply the changes and do some more testing.

msm-code:
> What's the usecase of requirements.plain.txt, should we maybe put it in a comment somewhere?

The idea was that I had to do some guesswork to find "original" dependencies. As you can see we had some "old" dependencies that were removed (for example, werkzeug), but stayed in the requirements because they were a transitive dependency of some project.

Basically the difference between packages.json and packeges.lock or something. I don't know a good place for a file like this - any ideas?

msm-code:
Actually this also fixes #100 - added to the description.

nazywam (approved):

PR #247, opened 2021-12-07
Work on the Progress Bar UI

### LINKED ISSUE #64: Show query duration
**Description**
I think that it will be nice to show query duration both on the Query page as well as on the "/recent" page. It will be helpful for the user, as with the time they'll get to know the best-practices by measuring querying durations.

**Action items**
- [x] show query duration for running queries on the Query page
- [ ] show query duration for running queries on the Recent page
- [ ] show query duration for finished queries  on the Recent page
- [x] show ETA for running query

### LINKED ISSUE #183: Show job progress correctly
**Feature Category**

- [ ] Correctness
- [ ] Performance
- [ ] Other (please explain)

**Describe the problem**

See #178 for my previous attempt at this problem.

There is no way to tell how many files from ursdabb will match the query before actually running it, and it can take between few seconds to many minutes (especially for queries for many files at once).  
When the ursadb query is very precise (for example, returns only a few files and all of them match the yara rule) this leads to weird situations where the query is "100% done" all the time, but still in processing.

I'll update this issue with my UI mockup proposotion soon

#178

### LINKED ISSUE #193: Inform the user that the db is working, before the first results appear
**Feature Category**

- [ ] Correctness
- [ ] Performance
- [ ] Other (please explain)

**Describe the problem**

When the user runs a comples yara rule, it can take a while for the first results to appear. This may look like the db is stuck.

We could add a banner like:

```
Querying database... Your results will appear shortly
```

(issue suggested by third party)

msm-code:

This is my attempt to solve many of the issues plaguing our progress bar, including:
#64
#183 
#193 

Unfortunately, I'm **very** bad at UIs, and these are primarily UI fixes.

1. I've changed `/recent` window a bit, to show query duration. I couldn't find a good place for it anywhere, so I've added it below the progress bar:

image

2. I've also changed how the progress bar works. The main problem (and there were at least two issues about it) is that it's hard to understand what's going on. "Under the hood" ursadb and yara work concurrently, and it's hard to show this on one progress bar.

First I've tried to solve this by adding more information (yay for 5 colors at once on one progress bar), but I saw the error of my ways, repented, and actually removed some information. See the result on the video below: 

https://user-images.githubusercontent.com/7026881/145097304-fdd50bd9-e3b0-4fee-a3fe-25dceadb14d9.mp4

The main addition is the spinner in the lower-left corner. It can be in one of the few states:
- `Collecting datasets...` - the query execution didn't start. Database doesn't even know how many datasets there are to query.
- `Searching for candidates X/Y (nn%)` - mquery is querying ursadb for candidates. Yara matching is done in the background, and is shown on the progress bar (scaled by the percent of queried datasets - for example 50/100 files tested with 1/2 datasets queried = estimated 25% progress).
- `Matching Yara X/Y (nn%)` - finally, the final step.

The new behaviour is superior in many ways. For example, when ursadb is slower than mquery it shows progress (more) correctly:

image

Instead of the old behaviour:

image

I'm not sure if it looks good though. But at least I think it's more understandable?

**Test plan**
Clone this branch and ensure that progress bar works as expected. 

**Closing issues**

fixes #64
fixes #183 
fixes #193


src/mqueryfront/src/components/QueryTimer.js
```diff
@@ -43,7 +43,7 @@ class QueryTimer extends Component {
 
         if (this.props.isFinished) {
             const duration = this.props.job.finished - this.props.job.submitted;
-            return <i>Duration: {this.getRenderTime(duration)}</i>;
+            return <i>Dutation: {this.getRenderTime(duration)}</i>;
```
ITAYC0HEN, line 46:
```suggestion
            return <i>Duration: {this.getRenderTime(duration)}</i>;
```

ITAYC0HEN:
Some notes:
1. Maybe the "Status: [something]" label is unnecessary because we have enough information to indicate the status (100% is done, not 100% is either processing or new, 0% can be considered "new")
2. Why do you need the spinner? I mean, the progress bar and spinner are showing that something is in progress

msm-code:
> Why do you need the spinner? I mean, the progress bar and spinner are showing that something is in progress

Sometimes (especially on HDDs and similar slower hardware, and when datasets are large) the db is stuck in the 0/0 state for some time (double digit number of seconds). For me it's obvious that the DB is working, but it may be not obvious for users.

The "spinning" element sends a clear (IMO) message that things are set in motion and that DB is working hard on giving some results.

Even worse, when the users becomes impatient, they may click "query" again, and accidentally create a second similar query. This is another issue (or non-issue, depending on PoV), but clearer UI should solve it.

But I'm not too attached to this particular spinner. Just want to be as clear as possible that the query is running, and not "stuck" when at 0/0.

msm-code:
> Maybe the "Status: [something]" label is unnecessary because we have enough information to indicate the status (100% is done, not 100% is either processing or new, 0% can be considered "new")

Agree, I think it has become redundant. I'll remove it.

msm-code:
https://user-images.githubusercontent.com/7026881/145272795-0f1f780d-a284-4193-a470-58a5e9530feb.mp4

Updated version. I've also moved matches to the left, and unified progress bar in the status page with the progess bar in the recents page (they are almost the same now).

nazywam (approved):

Issue #23, opened 2020-04-03
Implement user permission model


ITAYC0HEN:
**Description**

I believe that implementing user management would be a good decision. Maybe something that is based on mwdb user permission model. This will help to make mquery suitable for big organizations. 
For example, some queries are too sensitive to share even with other people in the same organization

Similarly to mwdb, this will allow the user to choose whether they want all the users to view the newly added job or to submit it as a personal rule, visible only for them (and maybe for the system admin). In this case, the recent jobs page will show by default only the user's jobs, and with a toggle turn-on, it will show all the publicly available jobs.

In addition, currently mquery parser the "author" attribute from the Yara. But the author signed on a Yara rule isn't necessarily the person who executes the rule in mquery. Say, if someone is executing a rule from the internet written by a stranger. Later, it will be hard for the user to find this job since they don't remember the name of the stranger.

msm-code:
I've been thinking about this for the past few days now.

I really don't add user management to mquery. This will only complicate simple deployments, and big companies that care already have their own user management systems.

I've discussed this with other mquery users, and oauth/openid seems to be a common denominator. What do you think about optionally reducing default permissions, and adding an `openid login` feature to grant additional permissions to selected users.

In other words, there will be option to restrict permissions of anonymous users in configs. More permissions can be granted to selected users logged in via OpenId.

Keep in mind that this will be an optional feature, so users that don't want to bother with OpenId won't have to.

In the unlikely case that a company using mquery doesn't already have an identity management server, it's easy to set up one with keycloak (https://www.keycloak.org/getting-started/getting-started-docker): `docker run -p 8080:8080 -e KEYCLOAK_USER=admin -e KEYCLOAK_PASSWORD=admin quay.io/keycloak/keycloak:15.0.2`.

Pinging @ITAYC0HEN as issue creator and @psrok1 as my guru on OpenID.

ITAYC0HEN:
If I understand what you said, it sounds good to me.
Just to verify, one - if they want to - can create their MQuery instance system like the following?
1. Nobody can access before logging in
2. Some logged in users would have limited permissions (i.e not all features\screens will be available)
3. Some logged in users will have all permissions (aka Admins)

Also, can a person limit a query and its results only to themselves? This scenario was mentioned in the original post

msm-code:
Yes.

My (unspecified as of now) plan involves having some permissions grouped into three sets for simplicity (names subject to change):

 - nobody - no permissions, can't access the system
[...]

PR #252, opened 2021-12-19
Feature/OIDC

msm-code:

Users are not implemented, as per #23

There is a "login" and "logout" button on the navbar:

image

image

They are visible only after configuring `openid_login_url` field in the database config (so it should be safe to merge).

They actually login the user with OIDC.

**Test plan**
1. Set up mquery as usual
2. Go to keystone panel (http://localhost:8080/), create a realm `myrealm`, add a user, and create a OIDC client: 

image

3. Go to `/config` in mquery, and set openid_auth_url to `http://localhost:8080/auth/realms/myrealm/protocol/openid-connect/token` and `openid_login_url` to `http://localhost:8080/auth/realms/myrealm/protocol/openid-connect/auth?client_id=app&response_type=code`

**Closing issues**

WIP for #23

### DIFF
--- a/src/mqueryfront/src/App.js
+++ b/src/mqueryfront/src/App.js
@@ -5,12 +5,21 @@ import QueryPage from "./query/QueryPage";
 import RecentPage from "./recent/RecentPage";
 import StatusPage from "./status/StatusPage";
 import ConfigPage from "./config/ConfigPage";
+import AuthPage from "./auth/AuthPage";
 import axios from "axios";
 import { API_URL } from "./config";
 import "./App.css";
 
+function parseJWT(token) {
+    const base64Url = token.split(".")[1];
+    const base64 = base64Url.replace("-", "+").replace("_", "/");
+    return JSON.parse(Buffer.from(base64, "base64").toString("binary"));
+}
+
 function App() {
     const [config, setConfig] = useState(null);
+    const rawToken = localStorage.getItem("rawToken");
+    const token = rawToken ? parseJWT(rawToken) : null;
 
     useEffect(() => {
         axios.get(`${API_URL}/server`).then((response) => {
@@ -18,29 +27,29 @@ function App() {
         });
     }, []);
 
+    const login = (rawToken) => {
+        localStorage.setItem("rawToken", rawToken);
+        window.location.href = "/";
+    };
+
+    const logout = () => {
+        localStorage.removeItem("rawToken");
+        window.location.href = "/";
+    };
+
     return (
         <div className="App">
-            <Navigation />
+            <Navigation session={token} config={config} logout={logout} />
             <Routes>
-                <Route exact path="/" element={<QueryPage config={config} />} />
-                <Route
-                    path="/query/:hash"
-                    element={<QueryPage config={config} />}
-                />
-                <Route
-                    exact
-                    path="/recent"
-                    element={<RecentPage config={config} />}
-                />
+                <Route exact path="/" element={<QueryPage />} />
+                <Route path="/query/:hash" element={<QueryPage />} />
+                <Route exact path="/recent" element={<RecentPage />} />
+                <Route exact path="/config" element={<ConfigPage />} />
+                <Route exact path="/status" element={<StatusPage />} />
                 <Route
                     exact
-                    path="/config"
-                    element={<ConfigPage config={config} />}
-                />
-                <Route
-                    exact
-                    path="/status"
-                    element={<StatusPage config={config} />}
+                    path="/auth"
+                    element={<AuthPage config={config} login={login} />}
                 />
             </Routes>
         </div>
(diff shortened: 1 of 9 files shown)


src/mqueryfront/src/auth/AuthPage.js
```diff
@@ -0,0 +1,50 @@
...
+        if (this.props.config == prevProps.config || !this.props.config) {
+            return;
+        }
+        if (!this.props.config["openid_auth_url"]) {
+            this.setState({ error: "OIDC not configured" });
+        }
+        const queryString = window.location.search;
+        const urlParams = new URLSearchParams(queryString);
+        const code = urlParams.get("code");
+        const payload = `grant_type=authorization_code&code=${code}&client_id=app`;
```
nazywam, line 24: Kinda hacky tbh, we assume that we don't urlencode `code`

src/schema.py
```diff
@@ -127,3 +127,5 @@ class BackendStatusDatasetsSchema(BaseModel):
 
 class ServerSchema(BaseModel):
     version: str
+    openid_auth_url: str
```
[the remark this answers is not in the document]
msm-code: Great call! This is actually a bug further down in typing (`HGET` can return None), and would crash some instances if someone updated their instance carelessly (because this request is now done from every subpage).

Of course I plan to test everything before the release but still, nice to have a working master branch. I'll add tests to ensure things like this won't happen again.

src/mqueryfront/src/auth/AuthPage.js
psrok1: I agree, we can pass that payload via regular body or `params`>

msm-code: I think I've copied that code from example on the internet :seal:. It's the Internet's fault!

OTOH it's unusual today for API to get structured urlencoded POST payloads (instead of JSON) so I get where people are coming from.

src/mqueryfront/src/App.js
```diff
@@ -5,42 +5,51 @@ import QueryPage from "./query/QueryPage";
...
+    const rawToken = localStorage.getItem("token");
+    const token = rawToken ? parseJWT(localStorage.getItem("token")) : null;
 
     useEffect(() => {
         axios.get(`${API_URL}/server`).then((response) => {
             setConfig(response.data);
         });
     }, []);
 
+    const login = (token) => {
```
psrok1, line 30: I don't like that name mismatch here.

In upper-scope variables `token` is parsed payload part, but in arguments `token` means whole JWT. Local-storage key is also `token` and meaning is the same as in argument.

I would change the naming to something like:
- `token` for JWT
- `session` for parsed payload part (as `session` is argument for navigation)

src/mqueryfront/src/App.js
```diff
@@ -5,42 +5,51 @@ import QueryPage from "./query/QueryPage";
...
-                    exact
-                    path="/recent"
-                    element={<RecentPage config={config} />}
-                />
-                <Route
-                    exact
-                    path="/config"
-                    element={<ConfigPage config={config} />}
-                />
+                <Route exact path="/" element={<QueryPage />} />
```
psrok1, line 44: I think these routes can be refactored to:
```suggestion
                <Route exact path="/">
                   <QueryPage />
                </Route>
```
But it's a subject for another PR.

psrok1 (commented):
I think it's the last iteration of change requests, other parts look good :+1:

src/mqueryfront/src/App.js
msm-code: Are you sure it's possible? I had to redesign this page like tihs after I had to upgrade for the new version of react-router. I don't see any mention of this syntax on the https://reactrouter.com/ page, and when I try to do this I get the following error:

image

src/mqueryfront/src/auth/AuthPage.js
msm-code: (fixed in the new version)

src/schema.py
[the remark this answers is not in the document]
msm-code: (fixed in the new versioN0

src/mqueryfront/src/App.js
msm-code: Good call! I hope to refactor this to another component, when I have time. So this all will be encapsulated in something like `<LoginBoundary />` or similar.

I've actually changed  `token` in the methods to `rawToken` instead. That's because `token` really is a (parsed) token. Session sounds like the whole state shared between client and server (which is not much right now and is basically the token, but can grow).

nazywam (approved):

src/mqueryfront/src/App.js
psrok1: Ok, I see now. In v5 they moved to `children` and in v6 they started use `element` property. Eh these JS frameworks, they change so often 😄

So I'm the one that is out of date.

https://github.com/remix-run/react-router/blob/main/docs/upgrading/v5.md

PR #255, opened 2021-12-22
Enforce user permissions server-side

msm-code:
Parse, verify, and enforce user permissions server-side.

This still has some rough edges that will be improved upon in the follow-up PRs, including at least:

- [ ] refactor the config handling
- [ ] hide UI elements that the current user has no permissions to
- [ ] add a validation logic to the config view
- [ ] (optional) improve the UI of the config view

And (possibly in the future)
- [ ] limit users only to specific tags
- [ ] hide queries done by another user from a logged in user
- [ ] allow read-only users (low-priority)

Part of #23

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -12,10 +12,13 @@ from zmq import Again
 from lib.yaraparse import parse_yara
 
 from util import mquery_version
-from db import Database, JobId, MQUERY_PLUGIN_NAME
+from db import Database, JobId
 from typing import Any, Callable, List, Union, Dict, Iterable, Optional
 import tempfile
 import zipfile
+import jwt
+import base64
+from cryptography.hazmat.primitives import serialization
 
 from schema import (
     JobsSchema,
@@ -40,7 +43,7 @@ app = FastAPI()
 
 
 class User:
-    def __init__(self, token: Optional[str]) -> None:
+    def __init__(self, token: Optional[Dict]) -> None:
         self.__token = token
 
     @property
@@ -49,18 +52,15 @@ class User:
 
     @property
     def name(self) -> str:
-        if self.is_anonymous:
+        if self.__token is None:
             return "anonymous"
-        # This feature is still WIP. TODO - token parsing and validating
-        return "username"
+        return self.__token.get("preferred_username", "unknown")
 
-    @property
-    def roles(self) -> List[str]:
-        if self.is_anonymous:
-            # Temporarily enable everything to everyone
-            return ["loggedin", "admin"]
-        # This feature is still WIP. TODO - user roles and configuration
-        return ["loggedin", "admin"]
+    def roles(self, client_id: Optional[str]) -> List[str]:
+        if self.__token is None:
+            return []
+        access = self.__token.get("resource_access", {})
+        return access.get(client_id, {}).get("roles", [])
 
 
 async def current_user(authorization: Optional[str] = Header(None)) -> User:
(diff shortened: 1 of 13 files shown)


src/mqueryfront/src/api.js
```diff
@@ -9,14 +9,21 @@ export function parseJWT(token) {
...
+    return axios
+        .request(path, {
+            method: method,
+            data: payload,
+            params: params,
+            headers: headers,
+        })
+        .catch((error) => {
+            if (error.response.status === 401) {
+                window.location = "/auth";
```
nazywam, line 23: This won't be annoying where a user tries to do some action they don't have permission to do?

src/db.py
```diff
@@ -323,9 +323,14 @@ def get_active_agents(self) -> Dict[str, AgentSpecSchema]:
     def get_core_config(self) -> Dict[str, str]:
         """Gets a list of configuration fields for the mquery core."""
         return {
+            # Autentication-related config
+            "auth_enabled": "Enable and force authentication for all users",
```
nazywam, line 327: Just making sure that this will be in fact str

src/app.py
```diff
@@ -71,7 +71,20 @@ async def current_user(authorization: Optional[str] = Header(None)) -> User:
...
+    secret = db.get_mquery_config_key("openid_secret")
+    if secret is None:
+        return User(None)
+
+    public_key = serialization.load_der_public_key(base64.b64decode(secret))
+    try:
+        token_json = jwt.decode(
+            token, public_key, algorithms=["RS256"], audience="account"  # type: ignore
+        )
+    except jwt.ExpiredSignatureError:
```
nazywam, line 83: What will happend when an invalid token is parsed? Should we handle it as well?

src/app.py
```diff
@@ -49,18 +52,15 @@ def is_anonymous(self) -> bool:
...
-        if self.is_anonymous:
-            # Temporarily enable everything to everyone
-            return ["loggedin", "admin"]
-        # This feature is still WIP. TODO - user roles and configuration
-        return ["loggedin", "admin"]
+    def roles(self, client_id: Optional[str]) -> List[str]:
+        if self.__token is None:
+            return []
+        access = self.__token.get("resource_access", {})
+        return access.get(client_id, {}).get("roles", [])
```
nazywam, line 63: not a :fan: of this - maybe just split it into one or two if-cases?

src/db.py
msm-code: Yes. Right now there's no support for anything other than string configuration.

Everything has to end up serialised to string in the end. But I plan to add a layer of abstraction over the db config, since it's clearly getting a bit complicated. It will also help with validation.

src/app.py
msm-code: Not sure. I'll experiment with unrolling it a bit (there are as much as three `get`s in this method)

src/app.py
msm-code: it will raise a 500 ISE exception. Not sure if we have to handle this (client will handle this correctly. It also shouldn't send requests like this in the first place).

This `except` is specifically to return `401`, because this will force a redirect to the login page for the web UI (and give the client a chance to reauthenticate)

src/mqueryfront/src/api.js
msm-code: Ideally the web UI will hide all the actions that the client is not allowed to do, so this shouldn't be annoying.

Also 401 is only when user is not logged in - when user is logged in, they get 403 which is just a regular red-error-bar error

src/app.py
msm-code: Consider this: https://devblogs.microsoft.com/python/idiomatic-python-eafp-versus-lbyl/

src/app.py
```diff
@@ -12,10 +12,13 @@
 from lib.yaraparse import parse_yara
 
 from util import mquery_version
-from db import Database, JobId, MQUERY_PLUGIN_NAME
+from db import Database, JobId
```
nazywam, line 15: We probably should think about import sorting :thinking:

nazywam (approved):

src/app.py
msm-code: makes sense, we can do it together with reformat

PR #259, opened 2021-12-30
Fix/config UI

msm-code:

Config UI is unusable when there are long values

image

There is a max field width.

image

Yes, I'm aware that this is a hack (max width in pixels). But after trying random things for an hour I have enough CSS for today. I have officially no idea how to fix that, and I believe the new behaviour is strictly better than the old one, hence this should be merged (and maybe improved in the future).

### DIFF
--- a/src/mqueryfront/src/config/ConfigEntries.js
+++ b/src/mqueryfront/src/config/ConfigEntries.js
@@ -114,7 +114,15 @@ class ConfigRow extends Component {
                 </td>
                 <td className="w-25">
                     <div className="d-flex">
-                        <div className="flex-grow-1" style={{"word-wrap": "break-word", "max-width": "800px"}}>{valueControl}</div>
+                        <div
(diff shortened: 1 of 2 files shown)


ITAYC0HEN (approved):
can be lived with. Improvement from the previous

src/mqueryfront/src/config/ConfigEntries.js
```diff
@@ -114,7 +114,7 @@ class ConfigRow extends Component {
                 </td>
                 <td className="w-25">
                     <div className="d-flex">
-                        <div className="flex-grow-1">{valueControl}</div>
+                        <div className="flex-grow-1" style={{"word-wrap": "break-word", "max-width": "800px"}}>{valueControl}</div>
```
nazywam, line 117: Maybe create an issue that this needs refactoring in the future

nazywam (approved):

PR #269, opened 2022-04-10
Improve parsing of string count expressions

msm-code:

The bug was discovered by running something equivalent to:
```
rule rule_name
{
    meta:
        description = "description"
        author = "author"
        date = "2022-04-05"
        reference = "reference"
        hash = "hash"
    strings:
        $name = "string"
    condition:
        #name >=3
}
```

`>=` operator was not supported well enough. I've also found many other cases of similar problems.

I fix some of the problems. There are still many ways to improve this (for example, handle `#x != 0`)

**Test plan**

I've tried to avoid any bugs, but due to the nature of the code (explicitly checking types of operators and parameters) it's bug-prone and there may be obscure combinations of operators that will be broken by this. I've taken moderate care to avoid this, but consulting https://yara.readthedocs.io/en/stable/writingrules.html for any edge cases may be helpful in review.

**Closing issues**

Submitted privately, no github issue.

### DIFF
--- a/src/lib/yaraparse.py
+++ b/src/lib/yaraparse.py
@@ -491,12 +491,14 @@ class RuleParseEngine:
         if isinstance(condition.right_operand, IntLiteralExpression):
             if condition.right_operand.value > 0:
                 return self.traverse(condition.left_operand)
+        return None
 
     def le_expr(self, condition: LeExpression) -> Optional[UrsaExpression]:
         """ From `y <= #a` we can deduce, that #a must occur at least once if y>0 """
         if isinstance(condition.left_operand, IntLiteralExpression):
             if condition.left_operand.value > 0:
                 return self.traverse(condition.right_operand)
+        return None
 
     def eq_expr(self, condition: EqExpression) -> Optional[UrsaExpression]:
         """ From `x == literal` and literal > 0 we can deduce that x must occur """
--- a/src/lib/yaraparse.py
+++ b/src/lib/yaraparse.py
@@ -9,6 +9,9 @@ from yaramod import (  # type: ignore
     AnyExpression,
     EqExpression,
     GtExpression,
+    GeExpression,
+    LtExpression,
+    LeExpression,
     IdExpression,
     IntLiteralExpression,
     OfExpression,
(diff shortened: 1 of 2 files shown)


src/lib/yaraparse.py
```diff
@@ -476,11 +479,37 @@ def of_expr(self, condition: OfExpression) -> Optional[UrsaExpression]:
             return None
 
     def gt_expr(self, condition: GtExpression) -> Optional[UrsaExpression]:
-        left = self.traverse(condition.left_operand)
-        right = self.traverse(condition.right_operand)
-        return left or right
+        """ From `#a > 3` we can deduce, that #a must occur at least once """
+        return self.traverse(condition.left_operand)
```
nazywam, line 483: It seems that both `#a > 3` and `3 < #a` are legal conditions, will checking just the left one work?

nazywam: nvm

src/lib/yaraparse.py
```diff
@@ -476,11 +479,37 @@ def of_expr(self, condition: OfExpression) -> Optional[UrsaExpression]:
...
+                return self.traverse(condition.right_operand)
 
     def eq_expr(self, condition: EqExpression) -> Optional[UrsaExpression]:
+        """ From `x = y` we can deduce, that x and y must both occur - unless:
+            - x = 0, or y = 0
+            - both x and y are count expressions (both can be zero)
+        """
+        if isinstance(condition.left_operand, IntLiteralExpression):
+            if isinstance(condition.right_operand, IntLiteralExpression):
+                if condition.left_operand == condition.right_operand == 0:
```
nazywam, line 508: shouldn't this be checked separately for both sides? i.e. 
```python
if isinstance(condition.left_operand, IntLiteralExpression):
    if condition.left_operand == 0:
        return None
if isinstance(condition.right_operand, IntLiteralExpression):
    if condition.right_operand == 0:
        return None
```
Right now it only check for `0 = 0`  which seems kinda bizarre :thinking:

src/tests/test_yaraparse.py
```diff
@@ -24,3 +23,115 @@ def test_literal_to_hex():
...
+    query = rule_to_query(
+        """
+    rule test {
+        strings:
+            $x = "test"
+        condition:
+            #x == 0
+    }"""
+    )
+    assert query == "{74657374}"
```
nazywam, line 123: Shouldn't this eval to `{}`?

msm-code: Of course. It started failing after fixing the bug above. Thanks for noticing.

src/lib/yaraparse.py
msm-code: I've simplified this condition a lot. I wasn't confident enough that this is correct (what if yara adds another type of expression with another semantics?).

nazywam (approved):
Makes sense, I think? :thinking:

msm-code:
> Makes sense, I think? thinking

This was implemented directly to improve parsing of one of the yara rules, so it's not an arbitrary change. There are some public rules that instead of strings have only counts like `#someJump > 16`.

Will merge after rebasing onto #272

PR #271, opened 2022-04-12
Fix #270: Finish job with several parallel daemon groups

### LINKED ISSUE #270: Several UrsaDB backends setup in parallel: only single backend works
**Environment information**
* Mquery version (from the /status page): 1.2.0
* Ursadb version (from the /status page): 1.3.2+3797f9b
* Installation method:
  - [ ] Generic docker-compose
  - [x] Dev docker-compose
  - [ ] Native (from source)
  - [ ] Other (please explain)

**Reproduction Steps**

I want to use two parallel ursadb backends. So I wrote the following docker-compose file: [docker-compose source code](https://github.com/ekorolevanyrun/parallel-ursadb/blob/main/docker-compose.yml)

**Expected behaviour**

I expected that mquery-web requests all mquery daemons, waits responses from them and results with all responses merged.

**Actual behaviour the bug**

Every time I send request (from the mquery-web UI) I got the following error with one of ursadb backend:

```
[2022-04-08 13:42:24.863] [info] LOCK: 496fa68f
[2022-04-08 13:42:24.863] [info] TASK: start [4]: iterator "496fa68f" pop 10;
[2022-04-08 13:42:24.864] [error] Task 4 failed: tried to read non-existent iterator
[2022-04-08 13:42:24.864] [info] TASK: done [4] (in 1ms)
```

I assume that the first daemon deletes key `496fa68f` from the redis and the second daemon (and ursadb backend) can't finish the process because of the deleted redis-key.

**Screenshots**

I think no screenshots are needed

**Additional context**

For example, here is the log from another ursadb backend:

```
[2022-04-12 05:07:35.276] [info] TASK: start [3]: topology;
[2022-04-12 05:07:35.276] [info] TASK: done [3] (in 0ms)
```

ekorolevanyrun:

Job is never finishes if there are several daemon groups (docker-compose.yml for reproduce: https://github.com/ekorolevanyrun/parallel-ursadb/blob/main/docker-compose.yml)

The condition (https://github.com/ekorolevanyrun/mquery/blob/fix/issue_270/src/daemon.py#L275) will be truth only once when the latest daemon finishes the task.

**Test plan**
Before changes job is never finishes, after changes job finishes after the latest daemon finish the job.

**Closing issues**

fixes #270

### DIFF
--- a/src/daemon.py
+++ b/src/daemon.py
@@ -278,7 +278,7 @@ class Agent:
             and j.datasets_left == 0
         ):
             # The job is over, work of this agent as done.
-            self.db.agent_finish_job(job)
+            self.db.finish_job(job)
 
     def __process_task(self, task: AgentTask) -> None:
         """Dispatches and executes the next incoming task.
--- a/src/db.py
+++ b/src/db.py
@@ -294,6 +294,11 @@ class Database:
                 job.key, {"status": "done", "finished": int(time())}
             )
 
+    def finish_job(self, job: JobId) -> None:
+        self.redis.hmset(
+            job.key, {"status": "done", "finished": int(time())}
+        )
+
     def has_pending_search_tasks(self, agent_id: str, job: JobId) -> bool:
         return self.redis.llen(f"job-ds:{agent_id}:{job.hash}") == 0
 


msm-code:
Thanks for your PR! I'll try to reproduce this locally first, because the `agent_finish_job` was added precisely so job finishes when the last agent finishes his job.

But the code looks suspicious, so it's possible something changed in the meantime and stopped working as intended.

msm-code:
I think a better approach is #282 - it's simpler, and I think this version may fail with more than two agents (concurrency is hard). The gist is something like "all agents should report they're done (using agent_finish_job), and only then the whole query is complete. Sadly the state is easy to corrupt when anything goes wrong (i blame redis for our poor data model).

PR #284, opened 2022-09-04
Automatically run filter plugins on all downloaded files

### LINKED ISSUE #281: Use filter plugins when accessing files directly with the web interface
> Hey MSM! I think that filter-plugin can\should also be called when clicking the Download button (https://github.com/CERT-Polska/mquery/blob/fecbef082bcb4355abdab318b90ef0a6dc1b9330/src/app.py#L205), what do you think?
> So if my S3 bucket contains encrypted files, it is all nice when I can decrypt it prior of scanning (and indexing), but it's a problem when user downloads it encrypted

This observation is definitely correct, and the behaviour should be fixed.

msm-code:
This is a draft PR that fixes #281. There are some minor things missing, most importantly:
 - [x] document and/or explain the limitations
 - [x] add the cleanup code (without introducing race conditions)

Filter plugins are not exeuted when downloading files

Filter plugins are applied to downloaded files

**Test plan**
- enable archive plugin (`MQUERY_PLUGINS=plugins.archive:GzipPlugin`)
- index a .gz file
- run any query and download the file (it should be decompressed on the fly)

**Closing issues**

fixes #281

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -30,7 +30,7 @@ plugin, but to load your own plugin you need to create your own image.
 ## Filter plugins
 
 Filter plugins can be used to discard files quickly (before even running
-yara rules), or to process raw paths returned from Ursadb.
+yara rules), or to process paths returned from Ursadb.
 
 The very simple example of a filter plugin is
 [RegexBlacklistPlugin](https://github.com/CERT-Polska/mquery/tree/master/src/plugins/blacklist.py)
@@ -89,12 +89,18 @@ class GzipPlugin(MetadataPlugin):
 The same method can be used to, for example, automatically download and extract
 files from s3 automatically.
 
-One caveat worth calling out:
-Filter plugins are first executed in the context of daemon (when matching files),
-and later in the context of the web application (when downloading files from the UI).
-It's important that the set of filter plugins is the same in both cases.
-For example this means that all daemons must have the same plugins active.
-In most deployments this difference is not significant.
+Filter plugins are ran before yara matching, and before file downloads. To avoid
+unexpected behaviour, the same set of plugins should be active in the web UI and
+in the daemon.
+
+**Warning:** if you have multiple backends either ensure that all backends and the web
+frontend use the same set of plugins, or be very careful about how they interact.
+
+For example, imagine that of the backends does gzip decompression and the other doesn't.
+Without any filter plugins installed on the frontend, download results will contain a
+mix of compressed and uncompressed files. Right now the answer to this is to write a
+plugin that does conditional decompression depending on which backend a file came from.
+It's not handled automatically.
 
 ## Metadata plugins
 
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -89,6 +89,13 @@ class GzipPlugin(MetadataPlugin):
 The same method can be used to, for example, automatically download and extract
 files from s3 automatically.
 
+One caveat worth calling out:
+Filter plugins are first executed in the context of daemon (when matching files),
+and later in the context of the web application (when downloading files from the UI).
+It's important that the set of filter plugins is the same in both cases.
+For example this means that all daemons must have the same plugins active.
+In most deployments this difference is not significant.
+
 ## Metadata plugins
 
 Metadata plugins are used to enrich results with additional metadata.
--- a/src/plugins/__init__.py
+++ b/src/plugins/__init__.py
@@ -1,6 +1,8 @@
-from typing import List, Type
+from typing import List, Type, Optional
 from metadata import MetadataPlugin
 from importlib import import_module
+from db import Database
+import logging
 
 
 def parse_plugin_list(plugins: str) -> List[str]:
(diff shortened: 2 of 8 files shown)


ITAYC0HEN:
Nice! Looking forward for this one :)

docs/plugins.md
```diff
@@ -89,6 +89,13 @@ class GzipPlugin(MetadataPlugin):
 The same method can be used to, for example, automatically download and extract
 files from s3 automatically.
 
+One caveat worth calling out:
+Filter plugins are first executed in the context of daemon (when matching files),
+and later in the context of the web application (when downloading files from the UI).
+It's important that the set of filter plugins is the same in both cases.
+For example this means that all daemons must have the same plugins active.
+In most deployments this difference is not significant.
```
nazywam, line 97: What's the exact meaning of `In most deployments this difference is not significant.` here?
If it's similar to "it's very unlikely that you'll have to worry about" then maybe use `It's unlikely to happen in most deployments but be careful.`?

src/plugins/__init__.py
```diff
@@ -33,3 +35,34 @@ def load_plugins(specs: List[str]) -> List[Type[MetadataPlugin]]:
...
+
+    def filter(self, orig_name: str) -> Optional[str]:
+        """ Runs all available filter plugins on the provided file.
+        Returns new file path, or None. User should call cleanup() later. """
+        current_path: Optional[str] = orig_name
+        for plugin in self.active_plugins:
+            if not plugin.is_filter:
+                continue
+            if current_path:
+                current_path = plugin.filter(orig_name, current_path)
```
nazywam, line 63: The semantics of this snippet is kinda, weird, maybe a early `break` or `return` is missing somewhere?
As far as i can tell, currently, `current_path` can become None and be passed to `plugin.filter`. At the same time it looks like the `plugin.filter` method expects a `str` and not `Optional[str]` (calling gzip.open on it and expecting a str in typehints) 🤔

nazywam (approved):

docs/plugins.md
msm-code: I was wondering how to write this myself.

Basically, I don't think there's a single deployment at earth right now where this matters. To run into this you would have to:
 - have two or more different ursadb backends for a single mquery instance (this is already very unlikely. I've used it and it was proposed for CERT, but not sure anybody does it)
 - the backends would have to use a different sets of plugins. For example, one ursadb backend uses gzip compresion and one doesn't.

Spinning up a setup like this requires a good understanding of all the components and how they interact. It doesn't just happen by accident, and someone who did this will have an easy time debugging issues arising from divergent plugins.

So basically, this is just an acknowledgement that "we know that using multiple backends with different plugins is not well supported". 

With that in mind - do you think I should reword the warning?

src/plugins/__init__.py
msm-code: yeah, you're right :thinking: I wonder why typings didn't catch the issue

msm-code: oh wait, you're not right. When `current_path` becomes `None` we won't  call `.filter` again, because `if current_path` will return False.

I don't remember what the reasoning was for structuring the code that way. I think we can simplify and just return early when current_path becomes None.

docs/plugins.md
nazywam: How about a simple `If your deployment doesn't have multiple ursadb backends (which is most likely the case) you don't have to worry about it`?

src/plugins/__init__.py
nazywam: looks better :+1:

docs/plugins.md
ITAYC0HEN: > - have two or more different ursadb backends for a single mquery instance (this is already very unlikely. I've used it and it was proposed for CERT, but not sure anybody does it).  
> - the backends would have to use a different sets of plugins. For example, one ursadb backend uses gzip compresion and one doesn't.


Exactly our case :)

ITAYC0HEN:
@yankovs Please have a one-last-look before it is merged

yankovs:
We've waited patiently for this, nice work :)

From what I understand `PluginManager` has reference to all plugins in the plugins directory, right? It seems ok for very simple usage but seems to general for more specific usage. For example, if you want the usage of some plugin exactly in the daemon filtering step but not at the download step.  

I see that a plugin already has an identity, a name, through `get_name`. Maybe the `filter` method of the manager could take a list of plugin names and filter based on it (very meta). So, in the `download` path it could be:  
```python
final_path = plugin.filter(file_path, [
    "plugin_A",
    "plugin_B"
])
```
If nothing is specified, iterate over all.  

Another option I can think about is to have a bool in the plugin config itself that specifies that it is a `download_plugin`, or a `search_plugin`, or a `megabeets_plugin`. Then an api endpoint could specify "I only want to use download plugins". But this seems to restrictive and a bad design choice overall (in case more plugin "types" are added in the future, a lot of time will be wasted adding it).

docs/plugins.md
msm-code: Haha, I should never say "probably nobody does X" when talking about a software.

The problem is that plugins are a bit "dumb". For example imagine that you have one backend with gzip plugin (all files are ungzipped) and one backend without. Mquery then issues a search request, and gets a list of matches (it's just a list of paths/file IDs):

```
/backendA/aaaaa
/backendB/bbbbb (gzipped)
/backendA/cccccc
/backendB/bbbbb (gzipped)
```

Mquery must somehow know what to do with gzipped files, and which files are gzipped. There are at least two solutions:
 - make the plugin writer handle this. For example write a "smart gzip" that passthroughs files with paths/IDs `/backendA/`, and decompresses files with `/backendB/`.
 - handle this in mquery. Keep track of which plugins were used by the backend during a job, and decide in mquery which plugins to run on the basis. This will be a bit more flexible, but requires some more backend changes.

msm-code:
> From what I understand PluginManager has reference to all plugins in the plugins directory, right? It seems ok for very simple usage but seems too general for more specific usage. For example, if you want the usage of some plugin exactly in the daemon filtering step but not at the download step.

I'll contradict directly what I wrote in documentation (looks like I need to update it), but you don't have to use the same plugins in the daemon and in the web UI - from the system's perspective they're completely separate.

For simplicity, assume that daemon and mquery UI are on different servers. So you can have some plugins into the daemon, and a different set of plugins in the web UI. For a specific example, you can leave plugin set empty in the web UI, and you will get the current behaviour. Or you can create a filter plugin that will encrypt all files before the download.

Is this enough or do you have even more advenced setups in mind?

docs/plugins.md
msm-code: What do you think about this version?

msm-code:
@yankovs what do you think? In general, I think this PR solves the issue it's supposed to (#281). If you have a more specific use case, please share so we can think about it a bit.

If I understand you correctly, the specific situation you've mentioned (using a different set of plugins during the download and searching) is covered too. Using different sets of plugins depending on the backend is more tricky (but possible with some workarounds and tradeoffs). If you can give a more specific example of an use-case maybe we can create an issue for it and hopefully improve the experience.

PR #292, opened 2022-12-04
Fix/token catch

### LINKED ISSUE #291: Handle token errors more gracefully
There are many cases when token decode may fail. For example, in my case I had old token stored in local storage, and jwt decode kept failing. We don't handle this in the client properly. I guess in the current user we should just catch jwt.InvalidTokenError instead of a specific error.

msm-code:

Currently it's possible to encounter problems when authentication is misconfigured, but disabled.

This PR should fix this, by early-exitting auth code when auth is disabled.

It introduces a small performance penalty, by reading auth_enabled redis key every request. At some point in future we should probably cache config per request.

**Test plan**
Ideally, run mquery, enable auth, make sure it works, disable auth, corrupt the keys and ensure mquery still works.
Otherwise, run mquery, set auth values (except auth_enable) to garbage and ensure it still works.
Finally, you can just trust me I tested it :^)

**Closing issues**

fixes #291

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -66,6 +66,10 @@ class User:
 
 
 async def current_user(authorization: Optional[str] = Header(None)) -> User:
+    auth_enabled = db.get_mquery_config_key("auth_enabled")
+    if not auth_enabled or auth_enabled == "false":
+        return User(None)
+
     if not authorization:
         return User(None)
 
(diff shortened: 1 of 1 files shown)


msm-code:
This PR needs to be rebased on master when bump to python 3.10 is finished

nazywam (approved):

PR #293, opened 2022-12-04
Fix documentation for config

### LINKED ISSUE #290: Update the config documentation
Documentation mentions that `auth_enabled` is to be changed to `yes`. With the new validation, it should be "true" or "false"

msm-code:
As explained in #290, the docs are misleading currently, because we only accept "true" as a boolean value, not "yes".

This PR fixes that and makes it explicit in the config key description.

**Closing issues**

fixes #290


msm-code:
We didn't, and user accounts are still explicitly "experimental" in the documentation.

The change is also not that breaking. Some PRs ago, the web interface was updated to only allow "true"/"false" instead of arbitrary input values (like "yes", "1", etc). The API and database will still gladly accept "yes" (in case anybody is mad enough to use it directly). And python will still work with old db configurations, because the boolean check is `if not val or val == "false"`.

But even if this was a breaking change, I hope we agree that we don't have to support all temporary bugs found in experimental features on the master branch and not included in any release.

PR #298, opened 2022-12-14
Add a feature for daemon forks

### LINKED ISSUE #296: Make it possible to run multiple daemons processes in a single container
**Feature Category**

- [ ] Correctness
- [ ] Performance
- [x] Other (please explain)

**Describe the problem**

Right now the demon is a single process, and the only way to spawn more demons is to create more containers/pods

This is a good solution in most cases, but sometimes having one pod/container with daemon per ursadb is simpler from devops point of view. We can add a flag like `--scale 4` to daemon, so that it's very easy to run multiprocessed daemon in a single container. 

There's nothing wrong with the current approach, but it can be a bit wasteful in some cases.

msm-code:

The only way to run multiple workers is to replicate at the level of containers

User can also pass a flag --scale X to run X processes in a single container

**Closing issues**

fixes #296

### DIFF
--- a/src/daemon.py
+++ b/src/daemon.py
@@ -331,7 +331,9 @@ def run_mquery(args, process_index):
     setup_logging()
 
     agent_group_id = args.group_id
-    logging.info("Agent [%s] running (process %s)...", agent_group_id, process_index)
+    logging.info(
+        "Agent [%s] running (process %s)...", agent_group_id, process_index
+    )
 
     db = Database(config.REDIS_HOST, config.REDIS_PORT)
     agent = Agent(agent_group_id, config.BACKEND, db)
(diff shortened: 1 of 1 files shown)


src/daemon.py
```diff
@@ -326,23 +327,52 @@ def main_loop(self) -> None:
             self.__process_task(task)
 
 
-def main() -> None:
-    """Spawns a new agent process. Use argv if you want to use a different
-    group_id (it's `default` by default)
-    """
+def run_mquery(args, process_index):
```
nazywam, line 330:
```suggestion
def run_mquery(args: argparse.Namespace, process_index: int) -> None:
```

nazywam (approved):

PR #314, opened 2022-12-26
Create permissions for a locked down user role

### LINKED ISSUE #302: Create a limited user role
**Feature Category**

- [ ] Correctness
- [ ] Performance
- [x] Other (please explain)

**Describe the problem**

I think about setting up a publicly available instance. In the current implementation, even if current user is not an admin, they can cancel every query in the system.

When anonymous users are enabled, we may not want them to be able to cancel each other queries. We just want them to see their queries (and possibly queries of other people).

This is a case of a more general problem: how do we handle user permissions and roles. Some open-ended questions:

- We support anonymous browsing (i.e. anonymous user may be given role of an user). Should anonymous users be able to cancel their own searches? If yes - how do we identify them?
- Should logged-in users be able to cancel their queries? What about queries of other users?
- Should users see each other queries? The same goes - if we want every user to see only their queries, how do we implement this for anonymous users?

msm-code:

There is no way to give more fine-grained permissions - everyone is
either an admin or a regulra user.

As documented in docs/users.md, the user role now is split into four sub-roles

 - `can_view_queries`: Can view a query with a given ID, and matched files.
 - `can_manage_queries`: Can create, stop, and delete queries.
 - `can_list_queries`: Can list queries (for "recent jobs" tab).
 - `can_download_files`: Can download matched file contents.

That list of sub-permissions is not considered stable yet - in future PR
I may decide to rename, remove or add some of them.

**Test plan**

**Closing issues**

fixes #302

### DIFF
--- a/docs/users.md
+++ b/docs/users.md
@@ -1,7 +1,5 @@
 # User management
 
-**Warning: this is a WIP feature. Users are advised to wait for a 1.3 release before trying it out in production**
-
 By default, every mquery user enjoys full admin rights.
 This is useful for trying it out, and for locked down or
 small deployments. But for larger multi-user deplotments, it may
@@ -11,17 +9,30 @@ Optional user management in mquery is role-based, and handled by OIDC.
 
 ## Role-based permissions
 
-As of now (v1.3 pre-releases) there are two defined roles:
+There are two predefined permission sets that can be assigned to users:
 
+ - `admin`: has access to everything, including management features.
+ Can change the service configuration, manage datasets, etc.
+ Users with `admin` role can do everything `user`s can.
  - `user`: has access to the analyst features. Can
  create new search jobs, see and cancel every job, and download
- matched files.
- - `admin`: has access to the management features. Can change the
- service configuration, manage datasets, etc. Users with `admin`
- role can do everything `user`s can.
+ matched files. In the current version, users can see and browse
+ all jobs in the system.
 
-**Note**: Before the release, `user` role will be improved.
-Right now, normal user can see jobs belonging to other users.
+This role names are considered stable, and will continue to work in the future.
+
+User permissions are then split in more fine-grained permissions:
+ - `can_view_queries`: Can view a query with a given ID, and matched files.
+ - `can_manage_queries`: Can create, stop, and delete queries.
+ - `can_list_queries`: Can list queries (for "recent jobs" tab).
+ - `can_download_files`: Can download matched file contents.
+
+It's possible to assign some of this roles directly (instead of giving the
+`user` role), but they're currently not considered stable and
+may change in some future new version.
+
+(**Note**: in the current version there is no isolation between users, and
+users can view/stop/delete each other queries. This may change in the future)
 
 ## OIDC integration
 
--- a/docs/users.md
+++ b/docs/users.md
@@ -23,7 +23,7 @@ Role names are considered stable, and will continue to work in the future.
 
 User permissions are then split into more fine-grained permissions:
 
- - `can_view_queries`: Can view a query with a given ID, and matched files.
+ - `can_view_queries`: Can view the query and matched files with a given query ID.
  - `can_manage_queries`: Can create, stop, and delete queries.
  - `can_list_queries`: Can list queries (for "recent jobs" tab).
  - `can_download_files`: Can download matched file contents.
--- a/docs/users.md
+++ b/docs/users.md
@@ -16,7 +16,7 @@ There are two predefined permission sets that can be assigned to users:
  Users with `admin` role can do everything `user`s can.
  - `user`: has access to the analyst features. Can
  create new search jobs, see and cancel every job, and download
- matched files. In the current version, users can see and browse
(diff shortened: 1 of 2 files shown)


docs/users.md
```diff
@@ -11,17 +9,31 @@ Optional user management in mquery is role-based, and handled by OIDC.
...
+ - `admin`: has access to everything, including management features.
+ Can change the service configuration, manage datasets, etc.
+ Users with `admin` role can do everything `user`s can.
  - `user`: has access to the analyst features. Can
  create new search jobs, see and cancel every job, and download
- matched files.
- - `admin`: has access to the management features. Can change the
- service configuration, manage datasets, etc. Users with `admin`
- role can do everything `user`s can.
+ matched files. In the current version, users can see and browse
```
nazywam, line 19:
```suggestion
 matched files. In current version, users can see and browse
```

docs/users.md
```diff
@@ -11,17 +9,31 @@ Optional user management in mquery is role-based, and handled by OIDC.
...
+ all jobs in the system.
+
+This role names are considered stable, and will continue to work in the future.
```
nazywam, line 22:
```suggestion
Role names are considered stable, and will continue to work in the future.
```

docs/users.md
```diff
@@ -11,17 +9,31 @@ Optional user management in mquery is role-based, and handled by OIDC.
...
+
+User permissions are then split in more fine-grained permissions:
```
nazywam, line 24:
```suggestion
User permissions are then split into more fine-grained permissions:
```

nazywam, line 24:
```suggestion
User permissions are then split into more fine-grained permissions:
```

docs/users.md
```diff
@@ -11,17 +9,31 @@ Optional user management in mquery is role-based, and handled by OIDC.
...
+
+ - `can_view_queries`: Can view a query with a given ID, and matched files.
```
nazywam, line 26:
```suggestion
 - `can_view_queries`: Can view the query and matched files with a given query ID.
```

docs/users.md
```diff
@@ -11,17 +9,31 @@ Optional user management in mquery is role-based, and handled by OIDC.
...
+ - `can_manage_queries`: Can create, stop, and delete queries.
+ - `can_list_queries`: Can list queries (for "recent jobs" tab).
+ - `can_download_files`: Can download matched file contents.
+
+It's possible to assign some of this roles directly (instead of giving the
```
nazywam, line 31:
```suggestion
It's possible to assign some of these roles directly (instead of giving the
```

src/app.py
```diff
@@ -158,14 +158,14 @@ def __call__(self, user: User = Depends(current_user)):
             default_roles = [
                 role.strip() for role in auth_default_roles.split(",")
             ]
-        all_user_roles = list(set(user_roles + default_roles))
+        all_roles = list(set(user_roles + default_roles))
+        all_roles = sum((expand_role(role) for role in all_roles), [])
```
nazywam, line 162: Just making sure that the `sum` is here used on purpose? The usage looks super weird to me

src/app.py
```diff
@@ -174,8 +174,31 @@ def __call__(self, user: User = Depends(current_user)):
...
+# See docs/users.md for documentation on the permission model.
 is_admin = RoleChecker(["admin"])
 is_user = RoleChecker(["user"])
+can_view_queries = RoleChecker(["can_view_queries"])
+can_manage_queries = RoleChecker(["can_manage_queries"])
+can_list_queries = RoleChecker(["can_list_queries"])
+can_download_files = RoleChecker(["can_download_files"])
+
+
+def expand_role(role):
```
nazywam, line 186:
```suggestion
def expand_role(role: str) -> List[str]:
```

nazywam (approved):

src/app.py
msm-code: yes, this is basically concat (summing all the given lists). The idea is to get, well, list of all applicable permissions, instead a list of expanded permissions.

For example:

```
>>> test = [[1, 2, 3], [4, 5, 6], [7, 8]]
>>> sum(test, [])
[1, 2, 3, 4, 5, 6, 7, 8]
```

I'm not aware of a better idiom for concat in python. If that's unclear I can rewrite it to explicit for loop, or a bit complex list comprehension - I think this will work in this case too:

```
[perm for role in all_roles for perm in expand(role)]
```

src/app.py
```diff
@@ -236,14 +259,82 @@ def config_edit(data: RequestConfigEdit = Body(...)) -> StatusSchema:
...
 
+@app.get(
+    "/api/backend",
+    response_model=BackendStatusSchema,
+    tags=["internal"],
+    dependencies=[Depends(is_admin)],
+)
+def backend_status() -> BackendStatusSchema:
+    """
+    Returns the current status of backend services, and returns it. Intended to
```
yankovs, line 270:
```suggestion
    Gets the current status of backend services, and returns it. Intended to
```

Issue #297, opened 2022-12-13
Make it possible to limit the number of files checked with yara


msm-code:
**Feature Category**

- [ ] Correctness
- [x] User Interface / User Experience
- [x] Performance
- [ ] Other (please explain)

**Describe the problem**

Sometimes running a yara rule on millions of samples is not desirable. It may be better to inform the user that the yara rule is not suitable for mquery, and let them improve it.

**Describe the solution you'd like**

I think the best solution is to have a configurable max number of files to yara-match (let's say 1000).

Small caveat is that we don't know how many files will yara-match before we run ursadb query, but that shouldn't be a huge issue. But the question is, do we still want to yara match on that files, or just reject the query immediately (I'm not sure yet myself).

**Describe alternatives you've considered**

Implement some priorities and rate limits - so that slow queries can finish in their own time. I'm not sure if that's a good idea yet.

Issue #257, opened 2021-12-29
User model: Refactor the config handling


msm-code:
Part of #23 

Right now we handle config very roughly. For example let's take:

```python
auth_default_roles = db.get_mquery_config_key("auth_default_roles")
if auth_default_roles is None:
    default_roles = []
else:
    default_roles = [
        role.strip() for role in auth_default_roles.split(",")
    ]
```

1. Default value is not declared anywhere, just used ad-hoc
2. The code parses the config key at the point of usage. This obscures the intention and may lead to duplication
3. Last but not least, the value is not validated when changed by the user. And typos here may lock users out and even make the system unusable.

I think about something like:

```
default_roles = db.config.auth_default_roles 
```

Where the `auth_default_roles` is a property that does the magic (parsing and validation)

msm-cert:
in other words:

Change all instances of `db.get_mquery_config(...)` to something like `db.get_config` that returns a well-typed Python object.

michalkrzem:
Do I understand correctly, reading other threads on this topic, that the https://github.com/bwindsor/typed-config solution will not be applicable here due to incompatibility with Docker environments? Zgodnie z https://github.com/CERT-Polska/mquery/pull/324#issuecomment-1401286103 ?

michalkrzem:
What about solution like https://github.com/CERT-Polska/Artemis/blob/main/artemis/config.py

msm-cert:
Hi,

short version (as per PM), something like a class MqueryConfig with fields like this:

```
@property
def default_roles(self) -> List[UserRole]:
    auth_default_roles = db.get_mquery_config_key("auth_default_roles")
    if auth_default_roles is None:
        return []
    else:
        return [
            role.strip() for role in auth_default_roles.split(",")
        ]
```

(a typed property per wrapped config entry) will probably be OK.

>that the https://github.com/bwindsor/typed-config solution will not be applicable here due to incompatibility with Docker environments

In this case the problem is that typed-config is for config files whereas this is more dynamic (it's more like program settings, I guess?) I guess we *could* make a provider for typedconfig that would make it possible to use typedconfig for this, but I think it's an overkill

>What about solution like https://github.com/CERT-Polska/Artemis/blob/main/artemis/config.py
Interesting, I never saw Annotated[] using like this. This is similar to what I had in mind (see above), but we need dynamic dispatch because settings may change (so we need properties/functions instead of static functions).

PR #315, opened 2022-12-29
Make it possible to disallow slow queries (or warn user about the problem)

### LINKED ISSUE #308: Advise the user that the Yara query they're about to execute is inefficient
**Feature Category**

- [ ] Correctness
- [x] Performance
- [ ] Other (please explain)

**Describe the problem**

Currently, people who don't understand how mquery and mwdb work behind the scene can execute very bad yara queries and it'll be (a) expensive (b) inefficient (c) take forever to run.

Advise the user that the Yara query they're about to execute is inefficient. We can warn them and ask "Are you sure you want to proceed?" or we can decide to prevent inefficient yara altogether.

VT has a similar approach on their retrohunt.

msm-code:
- [ ] I've updated documentation to reflect my change (if applicable) - not yet, I'll add a separate issue for documentation update

#308 

>Currently, people who don't understand how mquery and mwdb work behind the scene can execute very bad yara queries and it'll be (a) expensive (b) inefficient (c) take forever to run.

image

If configuration option `query_disallow_degenerate` is set to "true", users won't be able to run any queries that would end up doing a whole backend scan. Of course they can still run very inefficient queries like `"aaa" | "bbb" | ccc" | "ddd"` or just `{00 00 00}`, so this won't help in every case, but that's all we can do at this stage (others will be handled by #297).

I know this is not exactly what #308 asked for. I wonder if anyone really wants to use the "soft" mode - warn only, don't reject the query. If not, I think it's good as just a configuration option.

**Closing issues**

fixes #308

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -18,7 +18,7 @@ from starlette.responses import Response, FileResponse, StreamingResponse  # typ
 from starlette.staticfiles import StaticFiles  # type: ignore
 from zmq import Again
 
-from lib.yaraparse import YaraRuleData, parse_yara
+from lib.yaraparse import parse_yara
 
 from util import mquery_version
 from db import Database, JobId
--- a/src/app.py
+++ b/src/app.py
@@ -18,7 +18,7 @@ from starlette.responses import Response, FileResponse, StreamingResponse  # typ
 from starlette.staticfiles import StaticFiles  # type: ignore
 from zmq import Again
 
-from lib.yaraparse import parse_yara
+from lib.yaraparse import YaraRuleData, parse_yara
 
 from util import mquery_version
 from db import Database, JobId
@@ -342,17 +342,25 @@ def query(
         ]
 
     degenerate_rules = [r.name for r in rules if r.parse().is_degenerate]
-    disallow_degenerate = db.get_mquery_config_key("query_disallow_degenerate")
-    if degenerate_rules and disallow_degenerate == "true":
+    if degenerate_rules and not data.force_slow_queries:
+        allow_degenerate = (
+            db.get_mquery_config_key("query_disallow_degenerate") != "true"
+        )
+        if allow_degenerate:
+            # Warning: "You can force a slow query" literal is used to
+            # pattern match on the error message in the frontend.
+            help_message = "You can force a slow query if you want."
+        else:
+            help_message = "This is not allowed by this server."
         degenerate_rule_names = ", ".join(degenerate_rules)
         doc_url = "https://cert-polska.github.io/mquery/docs/yara.html"
         raise HTTPException(
             status_code=400,
             detail=(
-                "Invalid query. "
-                "Some of the rules would require a Yara scan of every indexed "
-                "file, and this is not allowed by this instance. "
-                f"Problematic rules: {degenerate_rule_names}. "
+                "Invalid query. Some of the rules require a full Yara scan of"
+                "every indexed file. "
+                f"{help_message} "
+                f"Slow rules: {degenerate_rule_names}. "
                 f"Read {doc_url} for more details."
             ),
         )
--- a/src/app.py
+++ b/src/app.py
@@ -326,7 +326,7 @@ def query(
         )
 
     if not rules:
-        raise HTTPException(status_code=400, detail=f"No rule was specified.")
+        raise HTTPException(status_code=400, detail="No rule was specified.")
 
     if data.method == RequestQueryMethod.parse:
         return [
@@ -335,10 +335,23 @@ def query(
                 rule_author=rule.author,
                 is_global=rule.is_global,
                 is_private=rule.is_private,
+                is_degenerate=rule.parse().is_degenerate,
                 parsed=rule.parse().query,
             )
             for rule in rules
         ]
+
+    degenerate_rules = [r.name for r in rules if r.parse().is_degenerate]
+    if degenerate_rules:
+        degenerate_rule_names = ", ".join(degenerate_rules)
+        doc_url = "https://cert-polska.github.io/mquery/docs/yara.html"
+        raise HTTPException(status_code=400, detail=(
+            "Invalid query. "
+            "Some of the rules would require a Yara scan of every indexed "
+            "file, and this is not allowed by this instance. "
+            f"Problematic rules: {degenerate_rule_names}. "
+            f"Read {doc_url} for more details."
+        ))
 
     active_agents = db.get_active_agents()
 
--- a/src/app.py
+++ b/src/app.py
@@ -342,16 +342,20 @@ def query(
         ]
 
     degenerate_rules = [r.name for r in rules if r.parse().is_degenerate]
-    if degenerate_rules:
+    disallow_degenerate = db.get_mquery_config_key("query_disallow_degenerate")
+    if degenerate_rules and (disallow_degenerate == "true"):
         degenerate_rule_names = ", ".join(degenerate_rules)
         doc_url = "https://cert-polska.github.io/mquery/docs/yara.html"
-        raise HTTPException(status_code=400, detail=(
-            "Invalid query. "
-            "Some of the rules would require a Yara scan of every indexed "
-            "file, and this is not allowed by this instance. "
-            f"Problematic rules: {degenerate_rule_names}. "
-            f"Read {doc_url} for more details."
-        ))
+        raise HTTPException(
+            status_code=400,
+            detail=(
+                "Invalid query. "
+                "Some of the rules would require a Yara scan of every indexed "
+                "file, and this is not allowed by this instance. "
+                f"Problematic rules: {degenerate_rule_names}. "
+                f"Read {doc_url} for more details."
+            ),
+        )
 
     active_agents = db.get_active_agents()
 
--- a/src/app.py
+++ b/src/app.py
@@ -343,7 +343,7 @@ def query(
 
     degenerate_rules = [r.name for r in rules if r.parse().is_degenerate]
     disallow_degenerate = db.get_mquery_config_key("query_disallow_degenerate")
-    if degenerate_rules and (disallow_degenerate == "true"):
+    if degenerate_rules and disallow_degenerate == "true":
         degenerate_rule_names = ", ".join(degenerate_rules)
         doc_url = "https://cert-polska.github.io/mquery/docs/yara.html"
         raise HTTPException(
--- a/src/app.py
+++ b/src/app.py
@@ -342,11 +342,9 @@ def query(
         ]
 
     degenerate_rules = [r.name for r in rules if r.parse().is_degenerate]
-    if degenerate_rules and not data.force_slow_queries:
-        allow_slow_queries = (
-            db.get_mquery_config_key("query_allow_slow") == "true"
-        )
-        if allow_slow_queries:
+    allow_slow = db.get_mquery_config_key("query_allow_slow") == "true"
+    if degenerate_rules and not (allow_slow and data.force_slow_queries):
+        if allow_slow:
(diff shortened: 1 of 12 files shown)


src/app.py
```diff
@@ -335,11 +335,28 @@ def query(
...
+        degenerate_rule_names = ", ".join(degenerate_rules)
+        doc_url = "https://cert-polska.github.io/mquery/docs/yara.html"
+        raise HTTPException(
+            status_code=400,
+            detail=(
+                "Invalid query. "
+                "Some of the rules would require a Yara scan of every indexed "
+                "file, and this is not allowed by this instance. "
+                f"Problematic rules: {degenerate_rule_names}. "
+                f"Read {doc_url} for more details."
```
ITAYC0HEN, line 356: link is not clickable, I guess it should :)

ITAYC0HEN:
> I wonder if anyone really wants to use the "soft" mode - warn only

Sometimes users really want to just list all the files that has a certain [imphash](https://yara.readthedocs.io/en/latest/modules/pe.html#c.imphash) value, and it's very important for their research. (i hope this falls under "degenerate query")

```
pe.imphash() == "b8bb385806b89680e13fc0cf24f4431e"
``` 


On other times, users don't really understand the problem.


Combining warn-only mode, **and** #297 will make sure that even if the user really wants to proceed, they can and it will not be too bad for the system (thanks to the file-limit)

src/app.py
```diff
@@ -335,11 +335,28 @@ def query(
                 rule_author=rule.author,
                 is_global=rule.is_global,
                 is_private=rule.is_private,
+                is_degenerate=rule.parse().is_degenerate,
```
nazywam, line 338: We could probably call `rule.parse()` just once, somewhere above? In case it gets expensive in the future

src/app.py
```diff
@@ -335,11 +335,28 @@ def query(
...
                 parsed=rule.parse().query,
             )
             for rule in rules
         ]
 
+    degenerate_rules = [r.name for r in rules if r.parse().is_degenerate]
+    disallow_degenerate = db.get_mquery_config_key("query_disallow_degenerate")
+    if degenerate_rules and (disallow_degenerate == "true"):
```
nazywam, line 346: Seems like `disallow_degenerate == "true"` should be somewhere lower in the abstraction layer :thinking: (just a thought)

nazywam: Also, I don't think the parentheses are needed?
```suggestion
    if degenerate_rules and disallow_degenerate == "true":
```

nazywam (approved):

src/app.py
msm-code: Not possible with an error message (without severe hacks, like putting HTML code into a Python exception, or - even worse - parsing the error message on the frontend.

I guess I'll do it the harder way, and put the related alert logic into the frontend.

src/app.py
msm-code: Performance hit is negliglible, and even if it wasn't, it's already cached:

```python
    def parse(self) -> UrsaExpression:
        if self.__parsed is None:
            self.__parsed = self.__parse_internal()
        return self.__parsed
```

I worry more about readability. There's no way to assign it to a temporary variable in a list comprehension, but I can rewrite this all to an explicit for loop if you prefer :thinking: It would look like this:

```python
response_rules = []
for rule in rules:
    parsed_rule = rule.parse()
    response_rules.append(
        ParseResponseSchema(
            rule_name=rule.name,
            rule_author=rule.author,
            is_global=rule.is_global,
            is_private=rule.is_private,
            is_degenerate=parsed_rule.is_degenerate,
            parsed=parsed_rule.query,
        )
    )
return response_rules
```

I prefer the current version, but I'm not overly attached to it if you prefer this one.

msm-code:
> (i hope this falls under "degenerate query")

It does :). No viable ngrams == "degenerate".

> Combining warn-only mode, and https://github.com/CERT-Polska/mquery/issues/297 will make sure that even if the user really wants to proceed, they can and it will not be too bad for the system (thanks to the file-limit)

ACK, thanks for input. I'll rework the code then to just warn the user in the 'non-enforcing' mode.

src/app.py
msm-code: O, and of course you're right about the lower abstraction layer. It's planned and scheduled for the next release: #257.

The current config is a mess anyway, for example I've named it "disallow_degenerate" instead of "allow_degenerate", because empty value is the default, and I want the default to be allowed. Letting storage details leak into config key names is not the best abstraction.

ITAYC0HEN:
So 3 modes? Allow\Ignore, Warn, Prevent\Block?

ITAYC0HEN:
it can also be nice if file-limit can be configured to work:
1. always
2. only in warn and block modes
3. only in block mode
4. never

msm-code:
> So 3 modes? Allow\Ignore, Warn, Prevent\Block?

I was considering it, but I think just two modes - `Warn` and `Block` are enough. It will be very easy to bypass the warning, and even when we trust users absolutely it may be nice to get an information that rule is inefficient (this can for example happen when copy-pasting rules from the internet - sometimes they look good but have something like `or pe.imphash == '1234'` that ruins it for mquery. I have a WIP docs page about writing good yara rules).

So in short, I don't see the application for `Always` mode, but I'm not fundamentally against it if anyone needs it.

> it can also be nice if file-limit can be configured to work:

Haven't thought of that yet. I don't immediately like the idea, because:

- it adds a dependency between otherwise independent features)
- Degenerate (sorry for the name btw, best I could do) queries are not the only ones hard for mquery. I've tested with https://github.com/Neo23x0/signature-base on a dataset with 750k malicious files, and even though most of the rules returned way less than 1% of dataset to scan, there were some really bad (for mquery) rules. Especially top3 worst rules were "almost as bad" as degenerate ones: https://github.com/Neo23x0/signature-base/blob/master/yara/gen_excel_auto_open_evasion.yar (712806 files), https://github.com/Neo23x0/signature-base/blob/master/yara/crime_ransom_darkside.yar (695732 files) and https://github.com/Neo23x0/signature-base/blob/master/yara/gen_xor_hunting.yar (662770 files).

But this is probably discussion for #297

msm-code:
@nazywam @ITAYC0HEN can you re-review?

Changes since the last time:

- I've changed `query_disallow_degenerate` to `query_allow_slow`. Nitpicker in me doesn't like the new name (slow queries are still possible, just harder), but it's probably easier to understand
- If that config key is set to "false" or empty, database will reject the query outright, and link the docs (I've updated them too)
- If that config is set to "true", database will ask the user if they're sure, the query button will change to scary red, and the search will start if the user clicks it again:

image

Also I've sneakily fixed broken CSS in the dropdowns.

Things I didn't change:

- Link is still not clickable. After tackling that for a while, I think this would require quite a few changes in the error handling on both server and client (at least if we want to do it correctly). I don't feel it's important enough improvement to justify the number of changes (link is copy&pasteable so the user can easily copy it). Maybe we can make it a (low priority?) issue.
- As per my previous message, there's no "Allow/Ignore" mode, only "Warn" and "Prevent/Block"

src/app.py
```diff
@@ -335,11 +335,34 @@ def query(
...
                 is_global=rule.is_global,
                 is_private=rule.is_private,
+                is_degenerate=rule.parse().is_degenerate,
                 parsed=rule.parse().query,
             )
             for rule in rules
         ]
 
+    degenerate_rules = [r.name for r in rules if r.parse().is_degenerate]
+    allow_slow = db.get_mquery_config_key("query_allow_slow") == "true"
```
yankovs, line 345: Probably nitpicky, but should configuration really be in Redis? From my understanding Redis has a couple of responsibilities in this project: it's used as some sort of state management to keep the state of plugins, but also to store configuration, and there may be other uses I don't remember of the top of my head (like as a cache mechanism, possibly?)

Seems a bit cluttered. Maybe all config should be moved to a config file with a (simple) config parser implemented? Maybe something like in Karton. I quite like their [approach](https://github.com/CERT-Polska/karton/blob/master/karton/core/config.py#L152-L162) to loading booleans and such. Makes stuff like line 345 be as elegant as `config.get("query_allow_slow")` which would equate to 0 or 1

msm-code: I don't think that's nitpicky! Quite the opposite, I think that's way too large issue to fit into a pull request comment :).

> but should configuration really be in Redis

Definitely not. Nothing should be in Redis except job scheduling internals (and maybe cache) (see also #74). Sadly we have to live with Redis for now, because I don't want to introduce another database at this stage of the project.

> Maybe all config should be moved to a config file with a (simple) config parser implemented? (...) Makes stuff like line 345 be as elegant as (...)

There are actually three issues hidden in this simple question:

**1. Refactor plugin configuration handling to remove the boilerplate**

This is already an issue here: #257. This is especially pressing for mquery's config fields. 

**2. Refactor mquery's config handling using typedconfig or configparser**

You may not think of it like this, but mquery already has a config file - [`config.py`](https://github.com/CERT-Polska/mquery/blob/66fba1b51490a40ca6c2978f6eaf99f350013b9d/src/config.docker.py). The intention of this file is to be modified with the deployment-specific config (this is why it's not tracked in git, for example). Things got more complicated because of docker - to make it more idiomatic for use in containers, we have config.docker.py that just reads all config fields from the environment (this is probably how you configure mquery).

I'm not fond of this pattern, but the current config is extremely simple (just backend URL, redis URL, forgotten "job expiration minutes" field #318 and plugin configuration).
I have it in the back of my mind to migrate someday to typedconfig or configparser, but (right now) it's a low priority refactoring and I'm weary of breaking backward compatibility.

**3. Move mquery core config fields from `plugin:Mquery` in the database to the config file**

(note: You say "all config should be moved", but I assume you mean all "core" config from `plugin:Mquery`. I don't think it's feasible to move plugin config to config file.)

After reading this I was initially dismissive of the idea, thinking it's unfeasible. After thinking a bit though, looks like [all current config fields](https://github.com/CERT-Polska/mquery/blob/66fba1b51490a40ca6c2978f6eaf99f350013b9d/src/db.py#L248) are viable candidates for a config file. This may be a bit harder for config fields that will influence worker behaviour (like #297 will), but still doable.

What I like about this idea:

* Less reinventing the whell, and fewer lines of code. Config refactor (#1) solved for free.
* Better separation of concerns - redis for plugins configuration, config file for mquery's core and web config.

What I don't like about this idea:

* This will break backward compatibility and we already have a release with redis-based mquery config. I think I can stomach that since I think number of orgs that integrated mquery auth with their orgs is very limited right now :).
* Impossible to configure mquery at runtime. Not necessarily a problem, current configuration keys should not be modified lightly and probably would benefit from being less interactive.
* At this point mquery's config gets a bit complicated, so we should combined this with (#2). I definitely don't want to have three configuration sources (redis, config.py, mquery.ini), therein lies madness. This means more backward-incompatible changes.

So wrapping up, I think you're onto something, but it needs a bit more design and it's definitely out of scope for this PR 🙃. Would you mind creating a separate issue for that?

ITAYC0HEN:
LGTM overall. Didn't check, but I hope that this feature works smartly with API requests as well, in some intuitive way.
Or is it?

msm-code:
> Didn't check, but I hope that this feature works smartly with API requests as well, in some intuitive way. Or is it?

Don't know about smartly, but it works:

* The `/api/query/` API has one new optional parameter - `force_slow_queries` (default is false). It works as expected - without it slow queries will be rejected. With the parameter set to true slow queries will be accepted (if mquery's configuration allows it).
* One caveat is that mquery error responses are not particularly API-friendly - just a JSON with an error message (no error code or something similar). So the API client doesn't know (without parsing the error message) what is the error cause (slow yara rule, syntax error, real internal server error, etc).

PR #317, opened 2023-01-07
Use rq as a task scheduler for daemon.

msm-code:
Mquery demon is getting unmaintainable. Early daemon was a trivial piece of code, and using redis queues directly as a RPC mechanism was the easiest solution. Since then mquery got extended with (in
no particular order) batching, users, jobs, commands, task cancellations,
distributed agents, configuration, and more.

Case in point: subjectively (and I think my opinion matters in that regard) there are more bugs in 350 LoC daemon.py than in the whole rest of this repository combined. 

We can keep digging that hole, or do something about this. I don't plan to rewrite the backend to another database, but at least we should use a better task queue than rolling ad-hoc our own one.

So this PR is an early **draft** of my ideas in that direction. So far it looks very promising: hopefully we'll see big readability, stability and correctness improvements. There is ~~two~~ one expected regression so far:

* Ideally, we would like to run only one ursadb task simulatanously (so we don't overwhelm ursadb). There's no obvious way to do this with rq. I partially simulate this by creating dependencies between jobs, so at least there won't be two simultanous ursadb queries for the same job

Last thing: usually I'm strongly against any rewrites, but in this case there many arguments for it:

* *incremental improvements are better* - but there is no way to change the job scheduling framework gradually, we need to swallow the pill sometime or keep making the situation worse
* *rewrites take time* - but this is just a simple worker with 350 lines of code. It shouldn't be that complicated.
* *rewrites introduce more bugs* - this code is already very buggy, and barely held together with a duck tape. I'm pretty sure - even though I can't prove it - that there are many race conditions that would manifest if I ran the db with some number of workers and fast/slow enough disk. So the code after rewrite will - hopefully - not be worse than the original. And I already have experience with debuging it


src/config.docker.py
```diff
@@ -1,7 +1,6 @@
 import os
 from plugins import parse_plugin_list
 
-INDEX_DIR = "/mnt/samples"
```
msm-code, line 4: unused since forever?

src/schema.py
```diff
@@ -20,7 +20,6 @@ class JobSchema(BaseModel):
     files_in_progress: int
     total_files: int
     files_errored: int
-    iterator: Optional[str]
```
msm-code, line 23: Unused since long time ago (iterators are not a global property of job since distributed workers are allowed == long time ago).

msm-code:
Ok, I think this is ready for review.

A few notes:

* This took longer than I expected, but was more fun than I expected. Rq is pretty neat and now I'm more sure of my code.
* Originally I wanted to do absolutely no changes to app.py, just db.py and daemon.py (to stop myself from doing more changes than necessary for my goal). In the end I couldn't stop myself from small refactoring - I've changed the JobId object to just string alias, and removed its uses from app.py. I've also changed a few parameter names. That's all. (Maybe I have a poor impulse control)
* I've added some comments to unrelated functions in the db. I've also documented data stored in the redis.
* A lot of the code is copy&pasted, but it's not clear from the git diff.

Overall I'm happy with this change and I think we can merge it (after a proper review)

src/db.py
```diff
@@ -68,31 +54,33 @@ def __init__(self, redis_host: str, redis_port: int) -> None:
...
-        """Gets yara rule associated with job"""
-        return self.redis.hget(job.key, "raw_yara")
+    def __schedule(self, agent: str, task: Any, *args: Any) -> None:
+        """Schedules the task to agent group `agent` using rq."""
+        Queue(agent, connection=self.redis).enqueue(task, *args)
 
     def get_job_ids(self) -> List[JobId]:
         """Gets IDs of all jobs in the database"""
-        return [JobId(key) for key in self.redis.keys("job:*")]
+        return [key[4:] for key in self.redis.keys("job:*")]
```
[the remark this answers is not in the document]
nazywam: yeah, made this comment before figuring it out, that's why i resolved it

Issue #316, opened 2022-12-29
Update the documentation for v1.4.0


msm-code:
**Feature Category**

- [ ] Correctness
- [x] User Interface / User Experience
- [ ] Performance
- [x] Other (please explain)

**Describe the problem**

Update the documentation. Especially, but not limited to:

- [ ] components of mquery, and how they interact in multi-server setting.
- [ ] how to write good yara rules (and link that from error in #315).

PR #333, opened 2023-01-15
Improve yara support documentation

msm-code:

Current documentation about yara support is too low-level - it explains "why" for sufficiently motivated users, but doesn't explain in simple terms what actually works wel;.

I think this version works better as an introduction to yara support by mquery.

**Test plan**
n/a

**Closing issues**

Related to #316

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -7,6 +7,8 @@
     mquery. In simple cases it can be solved without leaving the web UI, but
     many things will require more advanced approach. Read this if you need to
     index a considerable number of files.
+- [How to write good yara rules](./goodyara.md): How to write yara rules that
+    will work well in mquery.
 - [Yara support and limitations](./yara.md): Explains how mquery
     accelerates queries, what will, and what won't work.
 - [Utility scripts](./utils): Mquery ships with a few useful scripts.
--- a/docs/goodyara.md
+++ b/docs/goodyara.md
@@ -74,8 +74,8 @@ There are three things checked here:
 - PE imphash is either "7a861cd9c495e1d950a43cb708a22985" or "566a7a4ef613a797389b570f8b4f79df".
 
 You probably see where this is going - all of these conditions are ignored.
-Mquery doesn't care about 2-byte fragments (or their location in the file), or file
-size, and can't compute imphash. Mquery won't be able to speed up this yara rule, and will end up running yara on every file in the dataset (slow!).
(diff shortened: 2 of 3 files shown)


docs/goodyara.md
```diff
@@ -0,0 +1,156 @@
...
+```
+
+There are three things checked here:
+
+- First two bytes are 0x4d and 0x5a
+- File is smaller than 1MB
+- PE imphash is either "7a861cd9c495e1d950a43cb708a22985" or "566a7a4ef613a797389b570f8b4f79df".
+
+You probably see where this is going - all of these conditions are ignored.
+Mquery doesn't care about 2-byte fragments (or their location in the file), or file
```
nazywam, line 77:
```suggestion
Mquery doesn't care about 2-byte fragments (or their location in the file), or file
```

not sure if this shouldn't be lowercase `mquery` as well?

nazywam (approved):
:naziwam:

docs/goodyara.md
```diff
@@ -0,0 +1,156 @@
...
+size, and can't compute imphash. Mquery won't be able to speed up this yara rule, and will end up running yara on every file in the dataset (slow!).
```
nazywam, line 78:
```suggestion
size, and can't compute imphash. mquery won't be able to speed up this yara rule, and will end up running yara on every file in the dataset (slow!).
```

Oh, this one as well

docs/README.md
```diff
@@ -7,6 +7,8 @@
     mquery. In simple cases it can be solved without leaving the web UI, but
     many things will require more advanced approach. Read this if you need to
     index a considerable number of files.
+- [How to write good yara rules](./goodyara.md): How to write yara rules that
```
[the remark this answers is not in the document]
msm-code: Welp, looks like you're right. Maybe I should update this in other places in the docs too.

docs/goodyara.md
msm-code: It's a start of a sentence - so gramatically speaking probably should be capitalised? I usually try to avoid starting a sentence with proper names to avoid this kind of dillema, but if it happens like in this case it probably should start with a capital letter?

msm-code:
Current documentation about yara support is too low-level - it explains "why" for sufficiently motivated users, but doesn't explain in simple terms what actually works wel;.

I think this version works better as an introduction to yara support by mquery.

PR #335, opened 2023-01-17
Fix the progress bar

### LINKED ISSUE #319: Improve the progress bar
**Environment information**
* Mquery version (from the /status page): 1.3.0
* Ursadb version (from the /status page): 1.5.0
* Installation method:
  - [ ] Generic docker-compose
  - [ ] Dev docker-compose
  - [ ] Native (from source)
  - [x] Other (please explain): Installed all components separately (test environment)

I'll allow myself to skip the template.

Current progress bar has multiple problems. In no particular order:

* If an error happens before datasets are collected, progress bar stays in the "collecting datasets" state (even though job is cancelled already)
* If an error happens before all agents are queried, progress bar stays incomplete forever (even though job is cancelled)
* Progress bar behaves in a non-obvious way (because of a complex logic I won't explain here). There actually two progress bars in one - one thing we track is how many ursadb datasets were queried, and the second is how many candidates from ursadb were scanned with yara. I still think we should show what actually happens, but in a more understandable way to the user.
* Don't double-count errored and processed files (when file errors out, we increment files_errored counter. But we always show all processed files as green. So for example if all files fail, we show 50% green 50% red progress bar)
* I would also like to see the number of files checked with yara (maybe also with some % of true positives, to let the user know how good their rule is)

I'm not sure yet what the perfect progress bar would look like, but one thing I want to make sure is: when the query is finished (status: cancelled, or errored, etc) the progress bar is at 100% and no spinners are shown. If the status is finished but not "done" (success), the progress bar should be all red.

msm-code:

Explained in issue #319

This is a mix PR between "features" and "bugfixes"

> If an error happens before datasets are collected, progress bar stays in the "collecting datasets" state (even though job is cancelled already)

This one is fixed.

> If an error happens before all agents are queried, progress bar stays incomplete forever (even though job is cancelled)

This one too.

> Progress bar behaves in a non-obvious way (because of a complex logic I won't explain here). There actually two progress bars in one - one thing we track is how many ursadb datasets were queried, and the second is how many candidates from ursadb were scanned with yara. I still think we should show what actually happens, but in a more understandable way to the user.

I improved (?) it by adding a more elaborate description. In some cases it can get as long as this:

image

Irrelevant parts are not shown (for example, when all datasets are queried and collected already, only "matching: " part is visible.

I'm not sure about including the part `Backends working: 1`. It may be confusing, but on the other hand it explains progress bars like this:

image

Without the number of backends working visible it's very uncler why the query is still in progress (this is because in this case there were 10 dead backends configured).

> Don't double-count errored and processed files (when file errors out, we increment files_errored counter. But we always show all processed files as green. So for example if all files fail, we show 50% green 50% red progress bar)

Fixed.

> I would also like to see the number of files checked with yara (maybe also with some % of true positives, to let the user know how good their rule is)

Another UI change I'm not sure about. When the query is done, and only in the query view, matches are shown next to a number of candidates:

image

This is a very useful feature for me (as a developer and maybe a poewr user), not sure about the others.

**Test plan**

**Closing issues**

fixes #319

### DIFF
--- a/src/mqueryfront/src/components/QueryProgressBar.js
+++ b/src/mqueryfront/src/components/QueryProgressBar.js
@@ -54,7 +54,6 @@ const QueryProgressBar = (props) => {
     const datasetPct = Math.round(datasetFrac * 100);
 
     // TODO: remove the "failed" status after merging #317
-    console.log(status);
     if (status == "cancelled" || status == "failed") {
         return finalProgressBar(job, "query cancelled", "bg-danger");
     }
--- a/src/mqueryfront/src/components/QueryProgressBar.js
+++ b/src/mqueryfront/src/components/QueryProgressBar.js
@@ -1,9 +1,36 @@
 import React from "react";
 import ActionCancel from "./ActionCancel";
 import QueryTimer from "./QueryTimer";
-import { isStatusFinished, getProgressBarClass } from "../utils";
+import { isStatusFinished } from "../utils";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import { faSpinner } from "@fortawesome/free-solid-svg-icons";
+
+const finalProgressBar = (job, text, cssBg) => (
+    <div>
+        <div className="progress">
+            <div
+                className={"progress-bar " + cssBg}
+                role="progressbar"
+                style={{ width: "100%" }}
+                data-toggle="tooltip"
+                title={text}
+            >
+                {text}
+            </div>
+        </div>
+        <div>
+            <div className="float-right">
+                <QueryTimer
+                    job={job}
+                    isFinished={true}
+                    duration={true}
+                    countDown={true}
+                />
+            </div>
+            <div className="clearfix"></div>
+        </div>
+    </div>
+);
 
 const QueryProgressBar = (props) => {
     const { job, compact, onCancel } = props;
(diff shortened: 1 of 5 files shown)


src/mqueryfront/src/components/QueryProgressBar.js
```diff
@@ -16,74 +43,81 @@ const QueryProgressBar = (props) => {
...
+        total_files ? Math.round((files * datasetFrac * 100) / total_files) : 0;
+
     const datasetsDone = total_datasets - datasets_left;
     const datasetFrac = total_datasets > 0 ? datasetsDone / total_datasets : 0;
     const datasetPct = Math.round(datasetFrac * 100);
 
-    const getPercentage = (files) =>
-        total_files ? Math.round((files * datasetFrac * 100) / total_files) : 0;
+    // TODO: remove the "failed" status after merging #317
+    console.log(status);
```
nazywam, line 57: Leftover log, probably safe to remove?

nazywam (approved):

src/mqueryfront/src/components/QueryProgressBar.js
msm-code: Oof. Looks like CI rule that was meant to catch this stopped working at some point. Added issue for myself: #337

ITAYC0HEN:
> This is a very useful feature for me (as a developer and maybe a poewr user), not sure about the others

Well, this is true for the entire thing. non power-users would only want the "Okay dude when are you going to finish this thing?!@$" (ETA) and the number of results.

msm-code:
By entire thing do you mean the progress description?

Possibly. I find it annoying when I don't know what's going on, and have no way to check. For example, I'd like to know if a query is slow because ursadb processing takes too long (== complicated rule) or because there are too many candidates (== not strict enough rule). I'm not even sure it's completely irrelevant for non-power user, since it's an implicit feedback about "why" the rule is slow.

I would be OK with not showing this information by default / hiding it a bit, but I didn't find any way to do this in a reasonable way. Maybe when hovering a mouse over something?

Other than the UI changes, there are three bugfixes in this PR, hopefully non-controversial.

PR #339, opened 2023-01-21
Make it possible to limit the number of yara-scanned files

### LINKED ISSUE #297: Make it possible to limit the number of files checked with yara

msm-code:

Issue #297

This is a draft of a solution. It won't work yet, because I want to use typedconfg here after it's merged.

**Test plan**

1. Change the mquery.yara_limit config variable to something other than the default 0
2. Run a query that would return significantly more results than this
3. Ensure that the query does **not** return successfully. Instead it is cancelled as soon as it crosses the configured threshold of checked samples:
image

**Closing issues**

fixes #297

### DIFF
--- a/src/config.py
+++ b/src/config.py
@@ -20,7 +20,7 @@ class MqueryConfig(Config):
     plugins = key(cast=str, required=False, default="")
     # Maximum number of yara-scanned files per query (0 means no limit).
     yara_limit = key(cast=int, required=False, default=0)
-    # Html code to be displayed on the about page. 
+    # Html code to be displayed on the about page.
     about = key(cast=str, required=False, default="")
 
 
--- a/src/config.py
+++ b/src/config.py
@@ -5,14 +5,21 @@ import os
 
 @section("redis")
 class RedisConfig(Config):
+    # Hostname of a configured redis instance.
     host = key(cast=str, required=False, default="localhost")
+    # Port of a configured redis instance.
     port = key(cast=int, required=False, default=6379)
 
 
 @section("mquery")
 class MqueryConfig(Config):
+    # URL to a UrsaDB instance.
     backend = key(cast=str, required=False, default="tcp://127.0.0.1:9281")
+    # List of plugin specifications separated by comma, for example
+    # "plugins.archive:GzipPlugin, plugins.custom:CustomPlugin"
     plugins = key(cast=str, required=False, default="")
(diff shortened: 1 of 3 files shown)


src/config.py
```diff
@@ -5,14 +5,21 @@
...
+    # Port of a configured redis instance.
     port = key(cast=int, required=False, default=6379)
 
 
 @section("mquery")
 class MqueryConfig(Config):
+    # URL to a UrsaDB instance.
     backend = key(cast=str, required=False, default="tcp://127.0.0.1:9281")
+    # List of plugin specifications separated by comma, for example
+    # "plugins.archive:GzipPlugin, plugins.custom:CustomPlugin"
```
nazywam, line 19:
```suggestion
    # "plugins.archive:GzipPlugin,plugins.custom:CustomPlugin"
```
I guess?

src/db.py
```diff
@@ -148,14 +148,15 @@ def agent_add_tasks_in_progress(
...
         self, job: JobId, processed: int, matched: int, errored: int
-    ) -> None:
-        """Update progress for the job. This will increment number of files processed
-        and matched, and if as a result all files are processed, will change the job
-        status to `done`"""
-        self.redis.hincrby(f"job:{job}", "files_processed", processed)
+    ) -> int:
+        """Updates progress for the job. This will increment numbers processed,
+        inprogress, errored and matched files.
+        This will return the number of processed files after the operation."""
```
nazywam, line 154:
```suggestion
        Returns the number of processed files after the operation."""
```

nazywam (approved):

src/config.py
msm-code: Both will work (spaces are stripped), I like it with space a bit more (looks prettier)

PR #340, opened 2023-01-22
Remove old status types

msm-code:

Code handles some old, currently unused statuses

Dead code is removed. One ugly thing is that we need a fallback for the old statuses still in the database, since there is no mechanism for data migration in Redis (maybe we can hand-roll our own someday, unless we somehow decide to go with #74)  

**Test plan**

**Closing issues**

### DIFF
--- a/src/db.py
+++ b/src/db.py
@@ -79,6 +79,11 @@ class Database:
     def get_job(self, job: JobId) -> JobSchema:
         """Retrieves a job from the database. Tries to fix corrupted objects"""
         data = self.redis.hgetall(f"job:{job}")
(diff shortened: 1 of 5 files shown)


nazywam:
`utils/mquery.py:        FINISHED_STATES = ["cancelled", "failed", "done", "removed"]` I guess we're leaving this one for compatibility?

nazywam (approved):

msm-code:
Nice find, I actually missed it.

Ehh, I don't see an important reason to keep this for compatibility. This script is supposed to be used on new samples only. So cleaning it up shouldn't break anything, and leaving old statuses may be confusing. Removed them.

PR #327, opened 2023-01-12
Open-source, document and explain the S3 plugin/integration

msm-code:
**Your checklist for this pull request**
- [x] I've read the [contributing guideline](https://github.com/CERT-Polska/mquery/blob/master/CONTRIBUTING.md).
- [ ] I've tested my changes by building and running mquery, and testing changed functionality (if applicable)
- [ ] I've added automated tests for my change (if applicable, optional)
- [ ] I've updated documentation to reflect my change (if applicable)

**What is the current behaviour?**

No public documentation about s3 support

**What is the new behaviour?**

S3 support is documented. S3 plugin is open-sourced. Documentation is better.

**Test plan**

Literally half of the PR is a test plan guide.

**Closing issues**

fixes #326

PR #342, opened 2023-01-22
Plugin requirements install in web dockerfile

yankovs:

The web dockerfile build fails in case some plugin uses a python module not present in vanilla mquery due to the use of `PluginManager` and plugin loading in `app.py`

Build is successful.

**Test plan**
Build it and see if operational. I've tested this change with a plugin that uses `boto3` and it seems to work.

### DIFF
--- a/deploy/docker/daemon.Dockerfile
+++ b/deploy/docker/daemon.Dockerfile
@@ -1,8 +1,11 @@
 FROM python:3.10
 
 RUN apt update; apt install -y cmake
+
 COPY "requirements.txt" "/tmp/requirements.txt"
 RUN pip install -r /tmp/requirements.txt
+
+# plugin requirements
 # requirements.txt is added because at least one file must exist
 COPY requirements.txt src/plugins/requirements-*.txt /tmp/
 RUN ls /tmp/requirements-*.txt | xargs -i,, pip --no-cache-dir install -r ,,
--- a/deploy/docker/daemon.Dockerfile
+++ b/deploy/docker/daemon.Dockerfile
@@ -4,7 +4,7 @@ RUN apt update; apt install -y cmake
 
 # mquery and plugin requirements
 COPY requirements.txt src/plugins/requirements-*.txt /tmp/
-RUN ls /tmp/requirements-*.txt | xargs -i,, pip --no-cache-dir install -r ,,
+RUN ls /tmp/requirements*.txt | xargs -i,, pip --no-cache-dir install -r ,,
 
 COPY "src/" "/app"
 RUN chmod +x "/app/daemon.py"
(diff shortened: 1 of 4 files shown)


msm-code:
Thanks for your PR!

> The web dockerfile build fails in case some plugin uses a python module not present in vanilla mquery due to the use of PluginManager and plugin loading in app.py

Oof, right.

Would you mind also fixing `dev.web.Dockerfile`? I see that we don't install plugin dependencies there too. And maybe also add the `# plugin requirements` line to dev.daemon.Dockerfile` (just to keep things consistent).

One more idea, what do you think about changing:

```dockerfile
COPY "requirements.txt" "/tmp/requirements.txt"
RUN pip install -r /tmp/requirements.txt
# requirements.txt is added because at least one file must exist
COPY requirements.txt src/plugins/requirements-*.txt /tmp/
RUN ls /tmp/requirements-*.txt | xargs -i,, pip --no-cache-dir install -r ,,
```

To:

```dockerfile
# Copy and install main requirements.txt and plugin requirements.
COPY requirements.txt src/plugins/requirements-*.txt /tmp/
RUN ls /tmp/requirements*.txt | xargs -i,, pip --no-cache-dir install -r ,,
```

I didn't test thoroughly, but they look equivalent and I don't see a point of splitting the requirements installation process into two separate steps.

> Build it and see if operational. I've tested this change with a plugin that uses boto3 and it seems to work.

Looks like I should add requirements-s3.txt file to #327 too. And probably some tests, to catch similar misconfigurations in the docker images in the future.

This means that every official docker image we build & distribute will have (usually unnecessary) dependencies, like mwdblib and now minio. I'm inclined to accept this cost, since cost/benefit of a few megabytes more is worth lesser complexity.

yankovs:
Just notices that the actual line is:  
```bash
RUN ls /tmp/requirements-*.txt | xargs -i,, pip --no-cache-dir install -r ,,
```  
So it'll install only files of the form `requirements-*`. The ways I see around this (all of which I think are just aesthetically different):  
- Keep as is, with 2 requirements-install blocks, with the same redundancy and comments telling the story
- Remove the `-` in the `ls`, listing all files of the form `requirements*.txt`
- Merge those two install blocks to one, something of the form:
```bash
COPY requirements.txt src/plugins/requirements-*.txt /tmp/
RUN pip install -r /tmp/requirements.txt
RUN ls /tmp/requirements-*.txt | xargs -i,, pip --no-cache-dir install -r ,,
```  

@msm-code I think it's best to just stick with one and use it, what's your favorite 🙃?

msm-code:
>  `Remove the - in the ls, listing all files of the form requirements*.txt`

Thought about that one (I've put it in my snippet suggestion, but didn't call it out), so `ls requirements*.txt`

> @msm-code I think it's best to just stick with one and use it, what's your favorite?

I prefer the one with less docker layers, so `RUN ls /tmp/requirements*.txt | xargs -i,, pip --no-cache-dir install -r ,,`. Your third option is a tiny bit more cacheable, but don't think it's important in this use case.

msm-code (approved):
Looks good

PR #352, opened 2023-02-01
rq task timeout in config

yankovs:

This whole PR is based on a `rq.timeouts.JobTimeoutException` error during YARA matching in one of my plugins:
```python
Traceback (most recent call last):
  File "/app/plugins/SomePlugin.py", line 106, in filter
    shutil.copyfileobj(f_in, f_out)
  File "/usr/local/lib/python3.10/shutil.py", line 198, in copyfileobj
    fdst_write(buf)
  File "/usr/local/lib/python3.10/site-packages/rq/timeouts.py", line 61, in handle_death_penalty
    raise self._exception('Task exceeded maximum timeout value '
rq.timeouts.JobTimeoutException: Task exceeded maximum timeout value (180 seconds)
```
Consequently, I get these at the worker:  
```python
INFO:rq.worker:Killed horse pid 6434
INFO:rq.worker:Killed horse pid 6435
WARNING:rq.worker:Moving job to FailedJobRegistry (work-horse terminated unexpectedly; waitpid returned None)
WARNING:rq.worker:Moving job to FailedJobRegistry (work-horse terminated unexpectedly; waitpid returned None)
```  
And the most troublesome part is that the query hangs in 99%, which makes me believe mquery currently lacks rq-level error handling; a way to cleanup and gracefully shutdown search in case of such errors.

This PR adds a way to control top-level queued task timeout in `__schedule` method. Because code still needs to be tested this is currently in draft.  
Also, might check on the "query hangs on 99%" part in this PR, or leave it as is with just the config value, we'll see 🦆  

**Test plan**

TBD

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -41,9 +41,7 @@ Currently, supported configuration keys are:
   example `plugins.archive:GzipPlugin, plugins.custom:CustomPlugin`)
 - `mquery.yara_limit`: Maximum number of yara-scanned files per query (0 means no limit).
 - `mquery.about`: HTML code to be displayed on the about page.
-- `rq.start_timeout`: Timeout value for task initiation (in seconds). The default is OK for most users.
-- `rq.ursadb_timeout`: Timeout value (in seconds) for a ursadb query. The default is OK for most users.
-- `rq.yara_timeout`: Timeout value (in seconds) for yara scanning a batch of files. The default is OK for most users.
+- `rq.job_timeout`: Timeout value for rq jobs. The default is OK for most users.
 
 ## Mquery plugin configuration
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -39,6 +39,11 @@ Currently, supported configuration keys are:
   `tcp://ursadb-server:9281`)
 - `mquery.plugins`: List of supported plugins, separated by commas (for
   example `plugins.archive:GzipPlugin, plugins.custom:CustomPlugin`)
+- `mquery.yara_limit`: Maximum number of yara-scanned files per query (0 means no limit).
+- `mquery.about`: HTML code to be displayed on the about page.
+- `rq.start_timeout`: Timeout value for task initiation (in seconds). The default is OK for most users.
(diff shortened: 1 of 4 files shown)


yankovs:
> What do you think about rq timeouts in general? I wasn't aware the default is so low - I mostly use local files, so I never hit the timeout.
> 
> Maybe we could change the default timeout to something much higher (for example, 10 minutes, or even 30 minutes)? In normal operation the timeouts should only happen when someone kills a worker without giving it a chance to stop cleanly - so shouldn't happen in normal circumstances.
> 
> My point being, it doesn't have to be configurable if longer timeouts are better for everyone. What do you think?

I agree. I think that if rq job timeout is reasonably big, and it still fails, at this point the user either deals with huge files or has a very heavy plugin or something similar and should check their own setup. This shouldn't normally happen at all

> What do you think about rq timeouts in general? I wasn't aware the default is so low - I mostly use local files, so I never hit the timeout.
> 
I still think that in general, such rq exceptions should be handled in the code itself. I didn't show it (my bad 🥲) but above this traceback we got:  
```
Unknown error (plugin?): xyz.dll
```   
but in this case the file itself doesn't matter for the exception, if not `xyz.dll` this would've happened on `uvw.exe` later. So this is not very indicative of the issue, because basically it is rq-level and not caused (at least directly) from a faulty plugin.  
I think that: https://python-rq.org/docs/exceptions/ is a nice start to understand how to manage graceful shutdown of queries in cases of timeouts, Redis connection errors etc

msm-code (approved):
PR looks good. OTOH as per code review maybe you prefer to increase the default timeout a lot (and maybe make the batch size configurable)?

msm-code (approved):

msm-code:
I'll take a liberty of merging it even though it's still a draft, because it looks complete and is approved since Feb. I hope it's OK.

msm-code:
Actually unfortunately I had to make it a bit more complicated :cry:. The timeouts are not inherited by children, and we have a chain of task spawns flask -> enqueue a few start_search tasks -> each enqueues a few query_ursadb tasks -> each enqueues a few run_yara_batch tasks. Every enqueue invocation need job_timeout parameter too.

I've pushed my suggested changes to the PR.

Issue #299, opened 2022-12-19
Update web dependencies


yankovs:
**Feature Category**

- [ ] Correctness
- [ ] User Interface / User Experience
- [ ] Performance
- [X ] Other (please explain)

**Describe the problem**

Packages are old, this causes our CI/CD deployment of mquery (web) to indicate a lot of warnings:
![image](https://user-images.githubusercontent.com/49753115/208456235-fc45462c-f2a8-47f3-844a-8c34aa3ff4c9.png)

**Describe the solution you'd like**

Upgrade these, check it works, move on and be happy with your life

**Describe alternatives you've considered**

msm-code:
Updating is easy, checking if it works and fixing the regressions is not. Last time we did this (#260) we later had two regressions (#263, #294) even though it was tested. And bumping is not a very high priority, since security surface is very low (most vulnerabilities reported by CI aren't) 

But of course agreed, we should bump the dependencies (and ideally, keep them updated in the future).

msm-code:
Just head's up: I plan to work on that, but recently mquery's development picked up, and I don't want to work on this when there are PRs pending (this will probably cause merge conflicts).

PR #358, opened 2023-03-16
Update dependencies

yankovs:

This PR consists of updates to urgent, vulnerable packages mquery uses. In more details, I've done the following:  
- Replace `react-js-pagination` with `replace-js-pagination`. Yeah, this one is kinda stupid..  
Just like it says in the [npmjs page](https://www.npmjs.com/package/replace-js-pagination), it is derived from the package currently used and from my testing has the same interface. It requires minimal code change and [just works](https://media.tenor.com/rkI1a8s2Z6QAAAAC/todd-howard-it-just-works.gif) from my testing.
- Replace `react-multiselect-checkboxes` with `react-select`. The former seems abandoned and not updated for very long, and it also says on its npmjs page that it's based on `react-select` (which is actively maintained). Like the bullet above, the currently used package is based on the latter so changing required minimal code changes (as seen in diff).
- Updated `axios` to newest version.
- Moved `react-scripts` to `devDependencies`. This is actually recommended by CRA team, they give multi-paragraph explanation as to why it should be done. A bit sus, but reasonable nonetheless
- Ran `npm audit fix` and let it do it's thing.
- Removed unused `merge` package

This makes it so `npm audit --omit=dev` finds 0 vulnerabilities.

**Test plan**

My test plan was as follows:
- Set up dev env
- Make multiple files in `/samples`
- Index each with a different taint
- Make multiple YARA queries, check results make sense:
  - One without any taint selected that should not give any results
  - One with some taints that should not give any results
  - One without any taint that should bring some results
  - One with some taints that should bring some results
- Make enough YARA queries so that there are multiple pages in `recent jobs` page in mquery. Make sure pagination works

**Closing issues**

fixes https://github.com/CERT-Polska/mquery/issues/299


nazywam (approved):
Looks okay to me  :+1:

msm-code (approved):
Hi!

Sorry I let this PR rot for so long - my sabbatical interfered with my ability to focus on mquery (even though I thought it won't). Now I'm happily employed I should be able to focus on this project more :).

Built and tested it - LGTM.

PR #366, opened 2024-01-24
Remove unnecessary copies in dockerfiles

msm-code:
I believe these copies are not necessary, since `./public` directory is not used for anything at all by mquery web app.


msm-code:
OK, I think I debugged why they are necessary :thinking:

msm-cert (approved):

PR #367, opened 2024-01-24
Early work on migration to postgres for persistent storage

msm-code:
OK, time to tackle #74. I think it's time to finally break backward compatibility and release mquery 2.0.

One major pain point with developing mquery is how cursed the database is - redis was initially used when mquery was not meant to have (almost) any persistent storage, and only handled job queues and coordination.

Today thanks to ~~feature creep~~ user requirements it does much more, but every new persistent entity type is a pain to add. Hence, I think right now is a good moment to do it.

This PR moves plugin configuration from redis to the database. It's not a very important part of the system, but it's the easiest place to start this transformation.


src/db.py
```diff
@@ -320,3 +346,12 @@ def cache_get(self, key: str, expire: int) -> Optional[str]:
 
     def cache_store(self, key: str, value: str, expire: int) -> None:
         self.redis.setex(f"cached:{key}", expire, value)
+
+
+def init_db():
```
nazywam, line 351:
```suggestion
def init_db() -> None:
```

src/models/configentry.py
```diff
@@ -0,0 +1,9 @@
+from sqlmodel import Field, SQLModel
+from typing import Union
+
+
+class ConfigEntry(SQLModel, table=True):
+    id: Union[int, None] = Field(default=None, primary_key=True)
```
nazywam, line 6:
```suggestion
    id: int | None = Field(default=None, primary_key=True)
```

It'd be cool to use the new typing hints if our python is new enough

nazywam (approved):
lgtm

src/models/configentry.py
msm-code: It's not yet (but we'll get there)

PR #368, opened 2024-02-04
Feature/postgres job

msm-code:

Job objects are kept in redis

Job objects are kept in postgres.

I'm consciously trying to keep everything very self-contained and don't want to change a single line outside of the Database abstraction. It's very useful to have this layer when turning around the whole storage layer. 

Related to #74

### DIFF
--- a/src/db.py
+++ b/src/db.py
@@ -7,9 +7,10 @@ import string
 from redis import StrictRedis
 from enum import Enum
 from rq import Queue  # type: ignore
-from sqlmodel import Session, SQLModel, create_engine, select, and_
+from sqlmodel import Session, SQLModel, create_engine, select, and_, update
 
 from .models.configentry import ConfigEntry
+from .models.job import Job
 from .schema import JobSchema, MatchesSchema, AgentSpecSchema, ConfigSchema
 from .config import app_config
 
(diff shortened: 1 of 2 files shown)


docs/redis.md
```diff
@@ -22,9 +24,9 @@ To connect to redis use `redis-cli`. For docker-compose use
 
 Redis command documentation is pretty good and available at https://redis.io/commands/.
 
-### Job objects (`job:*`)
+### Job table (`job`)
 
-Job object is a Hash represented by schema.JobSchema class.
+Jobs are stored in a Job table.
```
nazywam, line 29:
```suggestion
Jobs are stored in the `job` table.
```

src/app.py
```diff
@@ -492,11 +492,11 @@ def matches(
 
 @app.get(
     "/api/job/{job_id}",
-    response_model=JobSchema,
+    response_model=Job,
```
nazywam, line 495: We're fine with returning the whole raw from db, right? I.e. there's no sensitive stuff there that users shouldn't have access to

nazywam: (Because if that was the case, we could define a schema for reading the job object)

src/db.py
```diff
@@ -67,55 +68,36 @@ def __schedule(self, agent: str, task: Any, *args: Any) -> None:
...
+            return [j.id for j in jobs]
 
-    def cancel_job(self, job: JobId) -> None:
-        """Sets the job status to cancelled"""
-        self.redis.hmset(
-            f"job:{job}",
-            {"status": "cancelled", "finished": int(time())},
-        )
+    def cancel_job(self, job: JobId, error=None) -> None:
+        """Sets the job status to cancelled, with optional error message"""
```
nazywam, line 76: What happens to the error message though?

src/db.py
```diff
@@ -67,55 +68,36 @@ def __schedule(self, agent: str, task: Any, *args: Any) -> None:
...
-        )
+        with Session(self.engine) as session:
+            return session.exec(select(Job).where(Job.id == job)).one()
 
     def remove_query(self, job: JobId) -> None:
         """Sets the job status to removed"""
-        self.redis.hmset(f"job:{job}", {"status": "removed"})
+        with Session(self.engine) as session:
+            session.execute(
+                update(Job).where(Job.id == job).values(status="removed")
```
nazywam, line 98: Some kind of enum/consts for the query statuses at some point probably?

nazywam (approved):

src/app.py
msm-cert: Hmm, for postgres migration I had to define `internal_id` for the Job object. It's not really sensitive, but by design it's not necessary for anything.

src/db.py
```diff
@@ -190,26 +210,29 @@ def create_search_task(
...
-            "files_matched": 0,
-            "files_errored": 0,
-            "total_files": 0,
-            "agents_left": len(agents),
-            "datasets_left": 0,
-            "total_datasets": 0,
-            "taints": json.dumps(taints),
-        }
+        with Session(self.engine) as session:
+            obj = Job(
```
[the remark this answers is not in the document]
msm-cert: Good question :thinking: depends on which one is clearer. Having defaults in the model for things like files_in_progress makes sense. When have a proper database I consider splitting this into several smaller objects (to avoid one huge job object that counts everything).

src/db.py
msm-cert: Created an issue out of it: #370

PR #371, opened 2024-02-05
Move the match objects to postgres

msm-cert:

Match objects are stored in redis

Match objects are stored in postgres

A lot of this code is still suboptimal boilerplate (for example, instead of:

1. get a job from the db
2. add a match

We are doing:

1. get a job from the db
2. extract a job ID from that job and throw away the rest
3. get a job with that ID from the database (!)
4. add a match to that job

Yes, this is clearly wasteful and a relict from redis time.

This will be fixed in the refactors following complete migration to postgres (we are nearly there).

related to #74

### DIFF
--- a/src/db.py
+++ b/src/db.py
@@ -11,6 +11,7 @@ from sqlmodel import Session, SQLModel, create_engine, select, and_, update
 
 from .models.configentry import ConfigEntry
 from .models.job import Job
+from .models.match import Match
 from .schema import MatchesSchema, AgentSpecSchema, ConfigSchema
 from .config import app_config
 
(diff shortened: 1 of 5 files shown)


src/db.py
```diff
@@ -240,22 +236,15 @@ def create_search_task(
...
-            if "sha256" not in match["meta"]:
-                match["meta"]["sha256"] = {
-                    "display_text": "0" * 64,
-                    "hidden": True,
-                }
-        return MatchesSchema(job=self.get_job(job), matches=matches)
+        with Session(self.engine) as session:
+            job = self.__get_job(session, job_id)
+            if limit is None:
+                matches = job.matches[offset:]
```
nazywam, line 244: This seems like an easy way out but we should probably do the pagination on the db side, right?
In cases where we'll be getting thousands of results it actually makes sense to make it gotta go fast.

nazywam: Oh, it's a single array field. :hmmm:

nazywam: Probably not worth it then

nazywam (approved):

nazywam (approved):

src/db.py
msm-cert: Nah, actually currently job.matches is a backreference to a list of Match objects (and every match object has an unrelated `matches` array... Duh, didn't realise until now. I already planned to rename i to `matched_rules`, but not all at once). Yeah, it makes a lot of sense to do this on the DB side, I was just lazy.

I've created an issue for it: #372.

PR #386, opened 2024-02-22
Draft: AgentJob object

msm-cert:
Draft: when this is finished, we will store all metadata about jobs in the agent specific objects instead of one huge Job structure.


msm-cert:
I didn't feel like finishing this refactoring right now. Instead I've opted in for a simpler solution: https://github.com/CERT-Polska/mquery/pull/388

PR #387, opened 2024-05-21
Fix Default Types in s3Index

humpalum:

The default is not set when no types are given

Sets the default

**Test plan**
--> Run s3index with no --types. If it Indexes with all the types its good.
--> Run with --types and make sure its not getting overwritten.

**Closing issues**
-

### DIFF
--- a/src/utils/s3index.py
+++ b/src/utils/s3index.py
@@ -98,10 +98,10 @@ def main() -> None:
 
     args = parser.parse_args()
     types = list(set(args.types))
-    
+
     if not args.types:
         types = ["gram3", "text4", "wide8", "hash4"]
-        
+
     if args.workdir is None:
         logging.error("--workdir is a required parameter")
         return
--- a/src/utils/s3index.py
+++ b/src/utils/s3index.py
@@ -100,7 +100,7 @@ def main() -> None:
     types = list(set(args.types))
     
     if not args.types:
-            types = ["gram3", "text4", "wide8", "hash4"]
+        types = ["gram3", "text4", "wide8", "hash4"]
         
     if args.workdir is None:
         logging.error("--workdir is a required parameter")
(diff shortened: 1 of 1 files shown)


nazywam (approved):

src/utils/s3index.py
```diff
@@ -99,6 +99,9 @@ def main() -> None:
     args = parser.parse_args()
     types = list(set(args.types))
 
+    if not args.types:
+        types = ["gram3", "text4", "wide8", "hash4"]
```
msm-cert, line 103:
```suggestion
        raise RuntimeError("You must pick at least one index type (see --type)")
```

msm-cert (changes requested):
Hi!

Thanks for your PR. This behaviour was changed ~semi recently (earlier all index types were the default). I think that it's not a good idea to use all index types by default (bad for the future if a new index type appears, and also not everyone needs all index types - for example omitting hash4 is a space-time tradeoff which may make sense).

What do you think about my suggestion? I'll also fix the documentation.

If you strongly think that the current behaviour is wrong and the default should be to use all index types, we may rethink it though.

src/utils/s3index.py
humpalum: Is no index type also a valid index type? At least it was possible to create a DB like that...
Basically it would be a "Just scan all Files with YARA" index... i dont see a exact use case right now but i could see someone having one.

I would just like to make sure its behaviour  is in sync with the other "index.py"

Anyways, i'd be fine with an error message.

humpalum: We would need to change the function docs which explains the default and mark it as required i guess

msm-cert: `Is no index type also a valid index type?` It is, but it's a bit pointless (as you've noticed). It will keep a list of files and you can later add indexes to an existing dataset (`reindex ... with ...` if I remember correctly), but that's a very weird use case so I don't think we have to support it (99% users don't want that I think).

I'll update the docs and ensure other indexes are in sync after merging.

msm-cert (approved):
Looks good. Thanks!

PR #393, opened 2024-08-14
Fix bug when there are unnecessary brackets in yara rule strings

ekorolev:
- [-] I've added automated tests for my change (if applicable, optional)
- [-] I've updated documentation to reflect my change (if applicable)

try yara rule:
```
rule parse_exception_example {
    strings:
        $xor_key_size   = { ((BB)|(68))??020000} 
        $c2         = { FF FF 68 74 74 70 }
    condition:
        all of them
}
```
Mquery can't parse it

The new additional function deletes unnecessary brackets and parser works well

**Test plan**
Just send a query with the following yara rule:
```
rule parse_exception_example {
    strings:
        $xor_key_size   = { ((BB)|(68))??020000} 
        $c2         = { FF FF 68 74 74 70 }
    condition:
        all of them
}
```

### DIFF
--- a/src/lib/yaraparse.py
+++ b/src/lib/yaraparse.py
@@ -126,10 +126,52 @@ class YaraRuleData:
     def is_private(self) -> bool:
         return self.rule.is_private
 
+def remove_unnecessary_brackets(hex_str: str) -> str:    
+    # Function to check if brackets are necessary
+    def is_bracket_necessary(content: str) -> bool:
+        return '|' in content
+
+    # Process the string
+    result = []
+    bracket_content = ""
+    in_bracket = False
+    
+    for char in hex_str:
+        if char == '(':
+            if in_bracket:
+                bracket_content += char
+            else:
+                in_bracket = True
+        elif char == ')':
+            if in_bracket:
+                if is_bracket_necessary(bracket_content):
+                    result.append(f"({bracket_content})")
+                else:
+                    result.append(bracket_content)
+                bracket_content = ""
+                in_bracket = False
+            else:
+                result.append(char)
+        else:
+            if in_bracket:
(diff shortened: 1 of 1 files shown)


msm-cert:
Hi! Nice find.

Thanks for the PR. I *think* this should be possible to do using the current yara parsing library, and if not the PR should go there instead of making a workaround in mquery (I did one before). I'll take a look at it.

msm-cert:
Thanks for your PR again.

I think the problem is a bit more complex - nested alternatives are not supported at all. For example consider:

```
rule parse_exception_example {
    strings:
        $xor_key_size   = { aa bb cc (11|22 (33 ??|44 ?? ??)) dd ee ff}
    condition:
        all of them
}
```

The rhetorical question in the source code was spot on:
```python
# alternatives, are nested alternatives a thing?
```

The fix will be a bit more complex, I think we have to parse the hex string and generate ursadb query from it. The example above is particularly complex, because the optimal ursadb query is:

```
{aa bb cc} & ({bb cc 11} | (bb cc 22 & (cc 22 33 | cc 22 44))) & {dd ee ff}
```

And there's no obvious way to convert from one representation to another.

ekorolev:
> Looks like I was wrong. Currently we don't try to optimize hex alternatives at all.
> 
> It's probably not a big deal - I've checked with https://github.com/Neo23x0/signature-base rule collection, and there's not a single rule that would benefit from this optimization. So it's probably not overly important.
> 
> But this makes the fix much easier. I've implemented it in #404 - just a single line change (and some documentation & tests). I hope you don't mind if I close this PR in favor of #404.

The main thing is that the issue has been fixed, thank you a lot!

Issue #402, opened 2024-09-17
Initialize the database automatically when starting mquery


msm-cert:
Currently the installation procedure is:

```
docker compose up --scale daemon=3  # this will take a while
docker compose exec web python3 -m mquery.db
```

Web container should initialize the database (run migrations) automatically.

This should happen when the *web* container is started.

Also update the documentation (remove the initialization step when it's not needed).

msm-cert:
Some hints how to do it:

1. executing `mquery.db` is actually not the best solution here, just `alembic upgrade head` would be better. See teste2e:
```
      - name: run web with docker compose
        run: docker compose up --build -d web --wait
      - name: init the database
        run: docker compose exec -it -w /usr/src/app/src/ web alembic upgrade head
      - name: run the rest of the code
        run: docker compose up -d
```

2. But we want to remove this line altogether. You can google for something like "fastapi automatic alembic migrations" etc. One of the resutls: https://stackoverflow.com/questions/77170361/running-alembic-migrations-on-fastapi-startup

3. You can also take a look at how mwdb does it: 
https://github.com/CERT-Polska/mwdb-core/blob/d593f697ebf37e036e778641038d5613ed852a5f/docker/start.sh#L10
https://github.com/CERT-Polska/mwdb-core/blob/d593f697ebf37e036e778641038d5613ed852a5f/mwdb/cli/cli.py#L60
https://github.com/CERT-Polska/mwdb-core/blob/d593f697ebf37e036e778641038d5613ed852a5f/mwdb/cli/database.py#L57

But this is more complex than we need for our needs.

PR #412, opened 2024-10-01
Create proper enums for job statuses

### LINKED ISSUE #370: Create proper enums for Job statuses
Related to #368. Mentioned https://github.com/CERT-Polska/mquery/pull/368/files#r1478115887

Instead of saving a status string like "new" to a database, we should have a proper enum value there.

michalkrzem:

Status column in Job model exists as a string

Status column in Job model is as a JobStatus enum.


src/migrations/versions/cbbba858deb0_init.py
```diff
@@ -5,84 +5,110 @@
...
 down_revision = None
 branch_labels = None
 depends_on = None
 
+DATABASE_URL = app_config.database.url
+engine = sa.create_engine(DATABASE_URL)
+
 
 def upgrade() -> None:
-    op.create_table(
```
msm-cert, line 18: Why this change? We should not change older migrations in any case. (Please also don't reformat them, to keep unnecessary git changes to minimum)

src/migrations/versions/dbb81bd4d47f_add_jobagent.py
```diff
@@ -5,34 +5,41 @@
...
 depends_on = None
 
+DATABASE_URL = app_config.database.url
+engine = sa.create_engine(DATABASE_URL)
+
 
 def upgrade() -> None:
     # ### commands auto generated by Alembic - please adjust! ###
-    op.create_table(
-        "jobagent",
```
msm-cert, line 20: similarly, please don't change and reformat this file

INSTALL.md
```diff
@@ -21,6 +21,7 @@ mkdir samples
 vim .env
 docker compose up --scale daemon=3  # this will take a while
 docker compose exec web python3 -m mquery.db
+docker compose exec web python3 -m alembic upgrade head
```
msm-cert, line 24: Not sure about this. We want to reduce the number of additional commands to zero (#402), not add more of them.

src/migrations/versions/84796bdc2f77_add_jobstatus.py
```diff
@@ -0,0 +1,28 @@
...
+# revision identifiers, used by Alembic.
+revision = "84796bdc2f77"
+down_revision = "dbb81bd4d47f"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    op.execute(
+        "CREATE TYPE jobstatus AS ENUM ('done', 'new', 'cancelled', 'removed', 'processing');"
```
msm-cert, line 18: You don't need to know every possible type for this migration. Just change every value out of scope to `cancelled`and then convert to enum.

msm-cert: (But good point that we need to handle this correctly)

msm-cert (changes requested):

msm-cert:
By the way, I've created an issue for removing the "removed" status. #414 

It may be a good time to implement it (for you it's just a few lines of code now, but it'll require another migration if done separately). If that's OK, please also remove this status from code at this time.

INSTALL.md
msm-cert: If you don't want to fix this in this PR and want to work separately in #402, it's OK to leave it as it is now (i.e. don't change the documentation in this PR. It works for fresh installations, just doesn't run migrations automatically).

src/migrations/versions/cbbba858deb0_init.py
michalkrzem: Ok. I understand. I did this because the alembic upgrade head command in an existing environment causes errors that the table exists. So I check beforehand if the table exists.

src/migrations/versions/dbb81bd4d47f_add_jobagent.py
michalkrzem: Ok

INSTALL.md
michalkrzem: > Not sure about this. We want to reduce the number of additional commands to zero (#402), not add more of them.

All because of alembic, which does not automatically recognize the modification from varchar to enum. I am able to create an enum type in postgres from the model level, but alembic does not replace it (it still remains VARCHAR) - that's why my manual modifications in migration files.

I found a library https://pypi.org/project/alembic-postgresql-enum/, quite new, I have to test it, but the question is whether to enter something like that at all. Or maybe look for solutions in sqlalchemy dialects for postgresql

msm-cert: don't change the documentation. After #421 alembic will be run with web automatically anyway.

INSTALL.md
```diff
@@ -39,6 +40,7 @@ cd mquery
 vim .env
 docker compose -f docker-compose.dev.yml up  # this will take a while
 docker compose exec dev-web python3 -m mquery.db
+docker compose exec dev-web python3 -m alembic upgrade head
```
msm-cert, line 43: don't change the documentation. After #421 alembic will be run with web automatically anyway.

src/migrations/versions/0820dad794e6_delete_removed_status.py
```diff
@@ -0,0 +1,46 @@
...
+            )
+        ],
+        enum_values_to_rename=[],
+    )
+    # ### end Alembic commands ###
+
+
+def downgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.sync_enum_values(
```
msm-cert, line 35: shouldn't this drop the table instead?

src/migrations/versions/cbbba858deb0_init.py
```diff
@@ -21,7 +20,9 @@ def upgrade() -> None:
         sa.Column("active_plugins", sa.ARRAY(sa.String()), nullable=True),
         sa.Column("name", sqlmodel.sql.sqltypes.AutoString(), nullable=False),
         sa.Column(
-            "ursadb_url", sqlmodel.sql.sqltypes.AutoString(), nullable=False
+            "ursadb_url",
+            sqlmodel.sql.sqltypes.AutoString(),
+            nullable=False,
```
msm-cert, line 25: ideally this line should not change (this PR is completely unrelated to this file). Is there anything wong with the previous formatting?

src/migrations/versions/f552a27b4390_modify_job_status_type_to_enum.py
```diff
@@ -0,0 +1,45 @@
...
+        "public",
+        "jobstatus",
+        ["done", "new", "cancelled", "removed", "processing"],
+        [
+            TableReference(
+                table_schema="public", table_name="job", column_name="status"
+            )
+        ],
+        enum_values_to_rename=[],
+    )
```
msm-cert, line 31: why two migrations for a single change?

src/migrations/versions/f552a27b4390_modify_job_status_type_to_enum.py
```diff
@@ -0,0 +1,45 @@
...
```
msm-cert, line 45: this should probably drop the table (downgrade)

src/models/job.py
```diff
@@ -1,3 +1,6 @@
+import enum
+
+from sqlalchemy import Enum as PgEnum
```
msm-cert, line 3: why not just Enum? If you want to rename it, DbEnum would be more appropriate

src/mqueryfront/src/utils.js
```diff
@@ -16,6 +15,7 @@ export const openidLoginUrl = (config) => {
     if (config["openid_url"] === null || config["openid_client_id"] === null) {
         // Defensive programming - config keys can be null.
         return "#";
+        gi;
```
msm-cert, line 18: typo

src/db.py
```diff
@@ -87,7 +91,7 @@ def get_valid_jobs(self, username_filter: Optional[str]) -> List[Job]:
         with self.session() as session:
             query = (
                 select(Job)
-                .where(Job.status != "removed")
+                .where(Job.status != JobStatus.cancelled)
```
msm-cert, line 94: No need for this where - idea of #414 is to remove all jobs with this status from the database. So there are no more jobs that were removed and need to be filtered out.
```suggestion
```

src/db.py
```diff
@@ -98,7 +102,9 @@ def remove_query(self, job: JobId) -> None:
         """Sets the job status to removed."""
         with self.session() as session:
             session.execute(
-                update(Job).where(Job.id == job).values(status="removed")
+                update(Job)
+                .where(Job.id == job)
+                .values(status=JobStatus.cancelled)
```
msm-cert, line 107: Again, the idea of #414 is to drop the jobs from the database. So this should be delete instead of update.

src/migrations/versions/0820dad794e6_delete_removed_status.py
```diff
@@ -0,0 +1,46 @@
...
+
+# revision identifiers, used by Alembic.
+revision = "0820dad794e6"
+down_revision = "f552a27b4390"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    op.execute("UPDATE job SET status = 'cancelled' WHERE status = 'removed';")
```
msm-cert, line 17:
```suggestion
    op.execute("DELETE FROM job WHERE status = 'removed';")
```
actually this works too and is even better

src/migrations/versions/0820dad794e6_delete_removed_status.py
```diff
@@ -0,0 +1,46 @@
...
+        enum_values_to_rename=[],
+    )
+    # ### end Alembic commands ###
+
+
+def downgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.sync_enum_values(
+        "public",
+        "jobstatus",
```
msm-cert, line 37: actually I don't care about downgrates too much, and it's not even possible in this case. You can just raise NotImplementedError() here.

requirements.txt
```diff
@@ -32,3 +32,4 @@ uvicorn==0.30.6
 wrapt==1.16.0
 yara-python==4.5.1
 yaramod==3.23.0
+alembic_postgresql_enum
```
msm-cert, line 35: I am positive this can be done without any additional dependencies.

https://stackoverflow.com/questions/63461381/how-to-use-an-existing-sqlalchemy-enum-in-an-alembic-migration-postgres
https://stackoverflow.com/questions/37848815/sqlalchemy-postgresql-enum-does-not-create-type-on-db-migrate

suggest a solution with postgresql.ENUM (I remember doing something like that too).

msm-cert (changes requested):

src/migrations/versions/cbbba858deb0_init.py
michalkrzem: I had to modify this because black raise errors.

src/mqueryfront/src/utils.js
msm-cert: still typo

src/migrations/versions/f654d7e4fcc7_on_delete_cascade_for_jobagent_and_match.py
```diff
@@ -0,0 +1,55 @@
...
+Revision ID: f654d7e4fcc7
+Revises: 6b495d5a4855
+Create Date: 2024-10-17 07:16:34.262079
+"""
+from alembic import op
+
+
+# revision identifiers, used by Alembic.
+revision = "f654d7e4fcc7"
+down_revision = "6b495d5a4855"
```
msm-cert, line 11: The order of migrations is wrong. This will run after the attempted jobs removal.

I've tested it:

```
mquery-dev-web-1       | WARNING:  StatReload detected changes in 'tasks.py'. Reloading...
mquery-dev-web-1       | INFO:     Started server process [24]
mquery-dev-web-1       | INFO:     Waiting for application startup.
mquery-dev-web-1       | INFO  [alembic.runtime.migration] Context impl PostgresqlImpl.
mquery-dev-web-1       | INFO  [alembic.runtime.migration] Will assume transactional DDL.
mquery-dev-web-1       | INFO  [alembic.runtime.migration] Running upgrade dbb81bd4d47f -> 6b495d5a4855, add jobstatus
mquery-dev-web-1       | Revision ID: 6b495d5a4855
mquery-dev-web-1       | Revises: dbb81bd4d47f
mquery-dev-web-1       | Create Date: 2024-10-15 08:17:30.036531
mquery-postgres-1      | 2024-10-18 14:31:24.310 UTC [51] ERROR:  update or delete on table "job" violates foreign key constraint "match_job_id_fkey" on table "match"
mquery-postgres-1      | 2024-10-18 14:31:24.310 UTC [51] DETAIL:  Key (internal_id)=(34) is still referenced from table "match".
mquery-postgres-1      | 2024-10-18 14:31:24.310 UTC [51] STATEMENT:  DELETE FROM job WHERE status = 'removed';
```

requirements.txt
msm-cert: as far as I can see it's not used and can be removed? We definitely don't want a new dependency.

src/migrations/versions/6b495d5a4855_add_jobstatus.py
```diff
@@ -0,0 +1,43 @@
...
+
+from mquery.models.job import JobStatus  # type: ignore # noqa
+
+# revision identifiers, used by Alembic.
+revision = "6b495d5a4855"
+down_revision = "dbb81bd4d47f"
+branch_labels = None
+depends_on = None
+
+job_status = sa.Enum(JobStatus, name="jobstatus")
```
msm-cert, line 17: This is a bit niche issue, but it is error prone in case JobStatus changes in the future.

This should be something like:

```
sa.Enum("done", "new", "cancelled", "processing", name="jobstatus")
```

(I didn't test it, but I think it should work)

src/migrations/versions/6b495d5a4855_add_jobstatus.py
```diff
@@ -0,0 +1,43 @@
+"""add jobstatus
+Revision ID: 6b495d5a4855
+Revises: dbb81bd4d47f
+Create Date: 2024-10-15 08:17:30.036531
+"""
+from alembic import op
+import sqlalchemy as sa
+
+from mquery.models.job import JobStatus  # type: ignore # noqa
```
msm-cert, line 9: see below, and then this import is not necessary

src/migrations/versions/f654d7e4fcc7_on_delete_cascade_for_jobagent_and_match.py
msm-cert: Both migrations can be probably merged into one to fix this.

msm-cert (changes requested):

msm-cert:
(by the way, feel free to close resolved conversations)

src/migrations/versions/f654d7e4fcc7_on_delete_cascade_for_jobagent_and_match.py
msm-cert: Yeah, sounds good. A benefit of one big migration is that if something goes wrong then a rollback is performed (so the db is not stuck in an invalid state).

msm-cert (approved):
LGTM!

PR #415, opened 2024-10-02
Fix static file dir in the dev builds

msm-cert:
Bumping the requirements caused starlette to check for dir existence and complain when it doesn't exist. In dev builds, this dir doesn't exist, and therefore is causing crashes.


mickol34:
Checked both development version and prod, both work flawlessly and without throwing additional starlette errors, thank you for help with this issue!

mickol34 (approved):

PR #420, opened 2024-10-09
feat: new JobFile model to store UrsaDB iterator batch files

### LINKED ISSUE #381: Store job files in the database instead of ursadb iterators
That way we can stop (ab)using iterators (and maybe even deprecate them in ursadb - they're a bit problematic in case of failed jobs).

mickol34:

Currently query to UrsaDB returns iterator and it's length which is then passed to batcher and YARA parser to unpack.

Now iterator is immediately popped after querying and several JobFile objects are created which store batches of files to execute YARA on.

**Test plan**

App should work the same way for various queries and rules. In case of other system failure during executing YARA there should remain JobFile instances (since those are deleted upon successful YARA execution).

**Closing issues**

fixes #381

### DIFF
--- a/src/models/queryresult.py
+++ b/src/models/queryresult.py
@@ -1,0 +1,7 @@
+from sqlmodel import Field, SQLModel, ARRAY, Column, String
+from typing import List
+
+
+class QueryResult(SQLModel, table=True):
+  job_id: str = Field(foreign_key="job.internal_id", primary_key=True)
+  files: List[str] = Field(sa_column=Column(ARRAY(String)))
--- a/src/models/queryresult.py
+++ b/src/models/queryresult.py
@@ -1,7 +1,8 @@
 from sqlmodel import Field, SQLModel, ARRAY, Column, String
-from typing import List
+from typing import List, Union
 
 
 class QueryResult(SQLModel, table=True):
-  job_id: str = Field(foreign_key="job.internal_id", primary_key=True)
-  files: List[str] = Field(sa_column=Column(ARRAY(String)))
+    id: Union[int, None] = Field(default=None, primary_key=True)
+    job_id: Union[int, None] = Field(foreign_key="job.internal_id")
+    files: List[str] = Field(sa_column=Column(ARRAY(String)))
--- a/src/tasks.py
+++ b/src/tasks.py
@@ -241,10 +241,9 @@ def query_ursadb(job_id: JobId, dataset_id: str, ursadb_query: str) -> None:
         if "error" in result:
             raise RuntimeError(result["error"])
 
-        files = result["files"]
-        agent.db.add_queryresult(job.internal_id, files)
-
-        file_count = len(files)
+        file_count = result["file_count"]
+        iterator = result["iterator"]
+        logging.info(f"Iterator {iterator} contains {file_count} files")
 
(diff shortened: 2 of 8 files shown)


src/tasks.py
```diff
@@ -259,34 +262,36 @@ def query_ursadb(job_id: JobId, dataset_id: str, ursadb_query: str) -> None:
...
-                iterator,
-                batch_size,
-                job_timeout=app_config.rq.job_timeout,
+        ## 1. get batch_size first files from result
+        batch_files = result.files[0:batch_size]
+
+        ## 2. remove batch files from result
+        with agent.db.session() as session:
+            session.execute(
+                update(QueryResult).where(QueryResult.job_id == result.job_id).values(files=result.files[batch_size+1:])
```
msm-cert, line 287: Doesn't look safe with regard to parallelism/multithreading.

msm-cert: Also you're missing one file every time (should be `[batch_size:]` probably).

src/models/queryresult.py
```diff
@@ -0,0 +1,7 @@
+from sqlmodel import Field, SQLModel, ARRAY, Column, String
+from typing import List
+
+
+class QueryResult(SQLModel, table=True):
+  job_id: str = Field(foreign_key="job.internal_id", primary_key=True)
+  files: List[str] = Field(sa_column=Column(ARRAY(String)))
```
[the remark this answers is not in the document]
mickol34: So this means I should multiple objects for every file returned from ```.query()```?

msm-cert:
BTW. this fails because QueryResult table does not exist in the DB. You need to create alembic migration for it (see also recent PR about enum rework). I didn't re-review the rest of the code yet.

mickol34:
Referring to #381 I'm not sure if ```ursadb.pop()``` is dead code, if it's used in ```all_indexed_files``` and ```all_indexed_names``` functions. If there are any other cleanups and/or tests to add, please comment below for me to fix them.

msm-cert (changes requested):
Ok, I think we can continue with this but with some large changes:

1. Restore query to the previous version (so return an iterator)
2. When querying, use that iterator to immediately add all the files from the result to the database (call `pop` in a loop and insert to the database, no need to handle `was_locked` state)
3. I think QueryResult should be something like `JobFile (id: int, job_id: int, file: str)`, so workers don't have to get all the results at once (can just select with limit and offset)
3. Alternatively, insert multiple QueryResults into the database, each with a batch of files (this is easier to implement but is less flexible. But still OK probably)
4. Importantly, when adding tasks: `agent.queue.enqueue(...` don't add filenames there. That's because it's stored in redis (== ram) and can potentially be very large. Instead, store just offset or batch_id (depending on which option in the previous you picked)

The reasoning here is that we prefer to have files in the database instead of in the opaque and weird ursadb iterator, but we have to be careful to avoid keeping the entire set of files at once in memory. In most cases this set will be small, but we can't OOM if someone does a huge query (this was actually a problem in the past).

I realise this may sound a bit convoluted, feel free to contact me in case of any questions

msm-cert:
Hi!

I'm closing it because since we didn't decide to merge it for 4 months it probably won't be merged any time soon.

But I still think it's a good idea, so maybe in the future we'll reuse this PR.

PR #421, opened 2024-10-11
Initialize the database automatically when starting mquery #402

### LINKED ISSUE #402: Initialize the database automatically when starting mquery

MichalJura:

Currently we needed to initilize database by hand: `python3 -m mquery.db`.

Now the database initialization is automatic and it uses `alembic upgrade head` instad of `mquery.db` so it also applies migrations. Also if the database already exists it will stamp it and create `alembic_version` for future migrations to be applied via `upgrade head`

**Test plan**

For new instance:  Create new instance and it will initilize database automatically.
For existing database: Remove (unnecessary for dev containers)  web container and set it up again.
Then you can check if there is `alembic_version` in postgresql.

**Closing issues**

fixes #402

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -20,7 +20,6 @@ mkdir samples
 # expect files in ./samples directory, and keep index in ./index.
 vim .env
 docker compose up --scale daemon=3  # this will take a while
-docker compose exec web python3 -m mquery.db
 ```
 
 - Good for testing mquery and production deployments on a single server
@@ -38,7 +37,6 @@ cd mquery
 # expect files in ./samples directory, and keep index in ./index.
 vim .env
 docker compose -f docker-compose.dev.yml up  # this will take a while
-docker compose exec dev-web python3 -m mquery.db
 ```
 
 - Good for development - all file changes will be picked up automatically.
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,1 +1,2 @@
 graft src/mqueryfront/dist
+include src/alembic.ini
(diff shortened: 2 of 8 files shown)


msm-cert (commented):
This looks more complicated than I thought it would. I'll take a look at it and check if it can be simplified.

MichalJura:
> This looks more complicated than I thought it would. I'll take a look at it and check if it can be simplified.

The reason behind it is that we don't have `alembic_version` table.
So for new instances there is no problem as `alembic upgrade head` will just apply migrations and add `alembic_version` to the database.
But if there is already a database `alemibc head upgrade` will result in error in Postrgres.
We could simply check if the database is initilized and then pass, but the new migrations like  #412  would have to be applied manually. 
On the other hand we migth want to apply them manually(but I don't think we are adding some breaking changes to the database schema?).
But in that case we could remove the stamping part.

msm-cert:
>The reason behind it is that we don't have alembic_version table.

We should, otherwise how would previous migrations be applied?

If previous documentation was giving incorrect instructions that bypass alembic it's unfortunate, but we don't have to accomodate such weird cases (database was not a part of any released version so it's not a critical problem).

msm-cert:
I still believe the previous version was severely overcomplicated.

Also:
image

You can't chdir to some place on disk randomly (this place is not even guaranteed to exist, it would only work with docker versions). Everything should work based on package.

Pushed some of my suggestions/fixes.

PR #424, opened 2024-10-16
Refactor user roles

### LINKED ISSUE #351: Refactor the role names
Should we maybe use some kind of enum for `can_list_all_queries`?

_Originally posted by @nazywam in https://github.com/CERT-Polska/mquery/pull/350#discussion_r1091704734_

I agree, planned to do this from a while ago but I didn't want to mix feature PRs with refactoring PRs.

msm-cert:

User roles are represented as plain strings.

User roles have their own enum.

**Test plan**
No user-visible behaviour changes, so everything should just work.

**Closing issues**

fixes #351


src/app.py
```diff
@@ -66,6 +67,22 @@ def with_plugins() -> Iterable[PluginManager]:
         plugins.cleanup()
 
 
+# See docs/users.md for documentation on the permission model.
+class UserRole(Enum):
+    # "role groups", used to grant a collection of "action roles"
+    nobody = auto()  # no permissions granted
```
nazywam, line 73: I'd be a bit worried about using `auto()` if we were storing the values somewhere as reordering the class fields could lead to a user accidentally becoming an admin.
But since as far as I can tell the enum instances are ephemeral it's fine 🤔

nazywam (approved):
lgtm

src/app.py
msm-cert: I agree. I don't see a good alternative though (explicit integers are super-ugly, manual strings `nobody = "nobody"` are verbose, and I'm unsure about using StrEnum.

I've decided to just document this issue/caveat for now.

PR #430, opened 2024-10-22
Try to ignore migration errors

msm-cert:

On migration error process dies because of unhandled exception and becomes a zombie

On migration error a clear error is logged, and process tries to keep going.

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -20,6 +20,7 @@ from typing import Any, Callable, List, Union, Dict, Iterable, Optional
 import tempfile
 import zipfile
 import jwt
+import logging
 import base64
 from cryptography.hazmat.primitives import serialization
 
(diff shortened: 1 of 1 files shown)


src/app.py
```diff
@@ -49,7 +50,10 @@
 
 @asynccontextmanager
 async def lifespan(app: FastAPI):
-    db.alembic_upgrade()
+    try:
+        db.alembic_upgrade()
+    except Exception:
+        logging.exception("Failed to apply migrations. Trying to continue...")
```
MichalJura, line 56: I think we should log the Exception as well e.g. f"Failed to apply migrations due to: {e} Trying to continue...".

msm-cert: Sure! Though in this case I've already thought about this :sweat_smile:

This is handled automatically by `logging.exception`, along with nice benefits like stack trace:

image

MichalJura (approved):

PR #433, opened 2024-10-30
257 user model refactor the config handling

### LINKED ISSUE #257: User model: Refactor the config handling

michalkrzem:

db.get_mquery_config_key("default_role) when called returns different types

I created UserModelConfig with properties  like auth_default_roles, openid_client_id, query_allow_slow etc. and now i call db.config.default_role

**Test plan**

**Closing issues**

fixes #257

### DIFF
--- a/src/db.py
+++ b/src/db.py
@@ -72,9 +72,10 @@ class UserModelConfig:
             "auth_default_roles"
         )
         if auth_default_roles is None:
-            return []
-        else:
-            return [role.strip() for role in auth_default_roles.split(",")]
+            auth_default_roles = "admin"
+        return [
+            UserRole[role.strip()] for role in auth_default_roles.split(",")
+        ]
 
     @property
     def openid_client_id(self) -> str:
--- a/src/db.py
+++ b/src/db.py
@@ -44,6 +44,22 @@ class TaskType(Enum):
 JobId = str
 
 
+class UserModelConfig:
+    def __init__(self, db_instance):
+        self.db = db_instance
+
+    def __getattr__(self, role_name):
+        # auth_default_roles = self.db.get_mquery_config_key(role_name)
+        # if auth_default_roles is None:
+        #     return []
+        # else:
+        #     print(role.strip() for role in auth_default_roles.split(","))
+        #     return [
+        #         role.strip() for role in auth_default_roles.split(",")
+        #     ]
+        return self.db.get_mquery_config_key(role_name)
+
+
 class Database:
     def __init__(self, redis_host: str, redis_port: int) -> None:
         self.redis: Any = StrictRedis(
@@ -56,6 +72,10 @@ class Database:
         Queue(agent, connection=self.redis).enqueue(
             task, *args, job_timeout=app_config.rq.job_timeout
         )
+
+    @property
+    def config(self):
+        return UserModelConfig(self)
 
     @contextmanager
     def session(self):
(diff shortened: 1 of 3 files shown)


src/db.py
```diff
@@ -40,10 +40,67 @@ class TaskType(Enum):
...
+    @property
+    def query_allow_slow(self) -> bool:
+        return self.db.get_mquery_config_key("query_allow_slow") == "true"
+
+    @property
+    def auth_enabled(self) -> bool:
+        auth_enabled = self.db.get_mquery_config_key("auth_enabled")
+        if not auth_enabled or auth_enabled == "false":
+            return False
+        return True
```
msm-cert, line 93:
```suggestion
        return self.db.get_mquery_config_key("auth_enabled") == "true"
```
Tiny chance in behaviour, but consistent with query_allow_slow (and works the same way for "true" and "false")

src/db.py
```diff
@@ -40,10 +40,67 @@ class TaskType(Enum):
...
+            "auth_default_roles"
+        )
+        if auth_default_roles is None:
+            auth_default_roles = "admin"
+        return [
+            UserRole[role.strip()] for role in auth_default_roles.split(",")
+        ]
+
+    @property
+    def openid_client_id(self) -> str:
```
msm-cert, line 81:
```suggestion
    def openid_client_id(self) -> str | None:
```

msm-cert: (if this causes type errors somewhere else, callers should be fixed to handle the situation where this is null)

src/db.py
```diff
@@ -40,10 +40,67 @@ class TaskType(Enum):
...
+
+    @property
+    def auth_enabled(self) -> bool:
+        auth_enabled = self.db.get_mquery_config_key("auth_enabled")
+        if not auth_enabled or auth_enabled == "false":
+            return False
+        return True
+
+    @property
+    def openid_url(self) -> str:
```
msm-cert, line 96:
```suggestion
    def openid_url(self) -> str | None:
```

src/schema.py
```diff
@@ -101,7 +101,7 @@ class BackendStatusDatasetsSchema(BaseModel):
 
 class ServerSchema(BaseModel):
     version: str
-    auth_enabled: Optional[str]
+    auth_enabled: Optional[bool]
```
msm-cert, line 104: I suspect this will cause issues on the frontend (but you can check). Unfortunately, you should probably cast to str when creating ServerSchema.

msm-cert (changes requested):

src/db.py
michalkrzem: brilliant :)

src/db.py
michalkrzem: Done

src/db.py
michalkrzem: Done

msm-cert (approved):
LGTM

PR #434, opened 2024-10-30
Add a QueuedFile object

msm-cert:

A smaller version of #432.

This PR creates a "core" (database object) of the index queue that will be managed by ursadb. This will let us queue files for indexing, and index them with available plugins.

This PR doesn't add the "indexer" worker yet, since it's nontrivial and needs some careful design.


msm-cert:
On the other hand, this idea needs some design. It's possible that the feature is too powerful and I should just special-case indexing s3 and local disk (because that's what everyone uses).

michalkrzem (approved):

src/migrations/versions/85ed22b90d72_add_queuedfile.py
```diff
@@ -0,0 +1,35 @@
...
+revision = "85ed22b90d72"
+down_revision = "6b495d5a4855"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table(
+        "queuedfile",
```
MichalJura, line 21: Should we add something like "indexed" column so we know which files are already indexed?
It would easier the workflow and allow to delete indexed files easily.

msm-cert: not "indexed", but other files will be added. Indexed is not necessary, because indexed files will be removed from queue.

PR #435, opened 2024-10-30
Start docker container start order

### LINKED ISSUE #428: Fix the docker-compose order for new projects
Small nitpick, so we don't forget.

After cloning mquery, when starting docker compose, daemon will often start first, and because of it it will start running before the database is initialized and crash. This dependency should be resolved, and the docs should be updated to reflect that (and relevant best practices).

msm-cert:
Make daemon depend on web, so the workers don't start before database init is complete.

Fixes #428

### DIFF
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -18,19 +18,23 @@ services:
       dockerfile: deploy/docker/dev.web.Dockerfile
     ports:
     - "5000:5000"
-    links:
-    - redis
-    - postgres
+    depends_on:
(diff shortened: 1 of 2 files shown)


mickol34:
LGTM in terms of code itself. Checked locally both prod and dev builds on pruned docker system, both seem to work flawlessly. In regards to #428 is there a docs page where this changes actually should be mentioned (e.g. `INSTALL.md`)?

mickol34 (approved):

PR #436, opened 2024-11-04
feat: show match context #396

### LINKED ISSUE #396: Show the information about matches in the frontend.
Related to #395 and #38.

After hovering over the matched rule name, we should see the match context (one can take inspiration from a screenshot in #38, or implement something else).

image

mickol34:

Currently there's no feature to show match context in search results.

This PR introduces "Show match context" action to search results using newly introduced backend Match field.

**Test plan**
Query some YARA, look at results, toggle modal with match context and check styling in various cases.

**Closing issues**

fixes #396

### DIFF
--- a/src/mqueryfront/src/components/ActionShowMatchContext.js
+++ b/src/mqueryfront/src/components/ActionShowMatchContext.js
@@ -52,32 +52,48 @@ const ActionShowMatchContext = (props) => {
             </button>
         </div>
     );
-    // Buffer.from(rawData, 'base64');
 
     const tableRows = Object.keys(props.context).map((rulename, index) => {
-        const rulenameRows = props.context[rulename].map((foundSample) => {
-            return (
-                <>
-                    <td scope="row">
-                        {atob(foundSample["before"])}
-                        {<b>{atob(foundSample["matching"])}</b>}
-                        {atob(foundSample["after"])}
-                    </td>
-                    <td scope="row">
-                        {base64ToHex(foundSample["before"])}
-                        {<b>{base64ToHex(foundSample["matching"])}</b>}
-                        {base64ToHex(foundSample["after"])}
-                    </td>
-                </>
-            );
-        });
+        const rulenameRows = Object.keys(props.context[rulename]).map(
+            (identifier) => {
+                const foundSample = props.context[rulename][identifier];
+                return (
+                    <>
+                        <td scope="row" style={{ width: "10%" }}>
+                            <span className="badge rounded-pill bg-info ms-1 mt-1">
+                                {identifier}
+                            </span>
+                        </td>
+                        <td
+                            scope="row"
+                            className="text-monospace"
+                            style={{ width: "25%" }}
+                        >
+                            {atob(foundSample["before"])}
+                            {<b>{atob(foundSample["matching"])}</b>}
+                            {atob(foundSample["after"])}
+                        </td>
+                        <td
+                            scope="row"
+                            className="text-monospace"
+                            style={{ width: "50%" }}
+                        >
+                            {base64ToHex(foundSample["before"])}
+                            {<b>{base64ToHex(foundSample["matching"])}</b>}
+                            {base64ToHex(foundSample["after"])}
+                        </td>
+                    </>
+                );
+            }
+        );
 
         return (
             <>
                 <tr key={index}>
                     <td
                         scope="row fit-content"
-                        rowSpan={props.context[rulename].length}
+                        rowSpan={Object.keys(props.context[rulename]).length}
+                        style={{ width: "15%" }}
                     >
                         <span className="badge rounded-pill bg-primary ms-1 mt-1">
                             {rulename}
@@ -97,7 +113,7 @@ const ActionShowMatchContext = (props) => {
             {!Object.keys(props.context).length ? (
                 "No context available"
             ) : (
-                <table className="table">
+                <table className="table table-bordered">
                     <tbody>{tableRows}</tbody>
                 </table>
             )}
--- a/src/mqueryfront/src/components/ActionShowMatchContext.js
+++ b/src/mqueryfront/src/components/ActionShowMatchContext.js
@@ -24,6 +24,16 @@ const useClickOutside = (ref, callback) => {
     });
 };
 
+function base64ToHex(str64) {
+    return atob(str64)
+        .split("")
+        .map(function (aChar) {
+            return ("0" + aChar.charCodeAt(0).toString(16)).slice(-2);
+        })
+        .join("")
+        .toUpperCase(); // Per your example output
+}
+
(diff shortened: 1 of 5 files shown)


michalkrzem:
[Screencast from 14.11.2024 12:18:49.webm](https://github.com/user-attachments/assets/c1a48d72-63eb-463a-a9a3-63f122b4dee2)

mickol34:
New modals are draggable and index themselves properly in front of everything else. New data structure allows to present data like in attached photo:
image

msm-cert:
Looks good! (just looking at screenshots, didn't review the code yet)

Can you align the byte and text views? (such that text and byte lines contain the same data, just encoded differently).

Can you separate bytes with spaces? I mean `70 79 74 68 6f 6e 0a` instead of `707974686f6e0a`. Unless it looks bad for some reason.

Nitpick: it's customary to show bytes first, so could you reorder the "bytes" and "text" columns? 

Finally, a question just in case - did you test this with non-printable bytes? When displaying, you can probably replace every character < 32 and >= 127 with a dot `.`.

mickol34:
Current styling and implementation with new API:
image
Works after removing stub content, but it would be great to test some various cases with multiple identifiers, weird data, bigger contexts and also multiple screen sizes and settings.

msm-cert:
Looks good!

Some small comments:

Consider changing the close button to: (change class to `btn-close` and remove custom `x` icon inside)

image

Text still doesn't align properly:

image

This may be an issue with my browser, but since people use various browers it's better to be resilient. I think the easiest solution would be to wrap this manually (use `white-space: pre` or `<br>` elements)

src/mqueryfront/src/components/ActionShowMatchContext.js
```diff
@@ -0,0 +1,175 @@
...
+    return (
+        atob(str64)
+            .split("")
+            .map(function (aChar) {
+                if (32 <= aChar.charCodeAt(0) && aChar.charCodeAt(0) < 127) {
+                    return aChar;
+                }
+                return ".";
+            })
+            .join(" ") + " "
```
msm-cert, line 49: maybe remove the whitespace between letters? So `(" ")`. I understand that now spacing depends on this, but this should be solved anyway.

src/mqueryfront/src/components/ActionShowMatchContext.js
```diff
@@ -0,0 +1,175 @@
...
+    useEffect(() => {
+        document.addEventListener("click", handleClick);
+
+        return () => {
+            document.removeEventListener("click", handleClick);
+        };
+    });
+};
+
+function base64ToHex(str64) {
```
msm-cert, line 27: Nitpick: I disagree slightly with this abstraction. I think it would be better to just hex encoding function.

It won't even make the code longer:

`toHex(atob(str))` vs
`base64ToHex(str)`

src/mqueryfront/src/components/ActionShowMatchContext.js
```diff
@@ -0,0 +1,175 @@
...
+            .split("")
+            .map(function (aChar) {
+                return ("0" + aChar.charCodeAt(0).toString(16)).slice(-2);
+            })
+            .join(" ")
+            .toUpperCase() + " "
+    );
+}
+
+function base64ToUtf8(str64) {
```
msm-cert, line 39: same goes for this function. I suggest `sanitize` and

`sanitize(atob(str))` instead of
`base64ToUtf8(str)`

src/mqueryfront/src/components/ActionShowMatchContext.js
```diff
@@ -0,0 +1,175 @@
...
+            })
+            .join(" ") + " "
+    );
+}
+
+const ActionShowMatchContext = (props) => {
+    const ref = useRef(null);
+    const [showModal, setShowModal] = useState(false);
+    const [focus, setFocus] = useState(true);
+    useClickOutside(ref, () => setFocus(false));
```
msm-cert, line 57: Sorry if that's obvious, but what does this line of code (and `useClickOutside`) do? I don't see anything happening when clicking outside.

src/mqueryfront/src/components/ActionShowMatchContext.js
mickol34: Done, now there's no spaces in between bytes or chars.

src/mqueryfront/src/components/ActionShowMatchContext.js
mickol34: I also slightly disagree with this nitpick, but solely due to my preference of code being more explicit and self-documenting.

src/mqueryfront/src/components/ActionShowMatchContext.js
mickol34: As mentioned above, I think that `sanitize` is too vague. If there will be another use case for such decryption in the project, it would be easier to get to know what was it's original purpose.

src/mqueryfront/src/components/ActionShowMatchContext.js
mickol34: Yeah, that's on my end for not documenting this enough (therefore name has been changed to `useClickOutsideModal`. 

So, the purpose of this hook is to detect if user clicked outside of this particular Element, in this case opened modal. This logic was needed to manage focus between overlapped modals. Originally, these modals would be autoindexed by browser to have different yet constant z-indexes and there would be no way to bring one of them "to the front". Now, by detecting clicks outside of modal we can give one of them significantly higher z-index and take it away if other modal was clicked. Hopefully this demo will explain the difference.

[Screencast from 2025-01-02 13-10-14.webm](https://github.com/user-attachments/assets/aa6e802d-a847-4fbd-99c7-028bf7f8a26a)

src/mqueryfront/src/components/ActionShowMatchContext.js
msm-code: This time I was probably unclear. The usual hexdump format, that it would be nice to have:

image

The important parts are:

* (usually) spaces between bytes
* no spaces between characters (to make plaintext easy to read)
* 16 characters per line
* optionally hexadecimal addresses on the left (we don't have that)

So spaces between bytes are OK (they make bytes easier to read). And the really important thing is 16 bytes per line. It may sound like I'm nitpicking, but I'm not ;)

As a POC:

image

msm-code (changes requested):
only the one comment about a function name, and about modal width, and we're ready to merge

src/mqueryfront/src/components/ActionShowMatchContext.js
msm-code: I just noticed in the source that the name is already `base64ToSanitizedUtf8`. That's much better (though still utf8 is technically incorrect)

msm-code (approved):

msm-cert (approved):

PR #439, opened 2024-11-22
395 add the information about match context to the database

michalkrzem:

**Test plan**

**Closing issues**

fixes #issuenumber


src/tasks.py
```diff
@@ -140,6 +194,36 @@ def execute_yara(self, job: Job, files: List[str]) -> None:
...
             )
 
+    def get_match_context(
+        self, data: bytes, matches: List[yara.Match]
+    ) -> dict:
+        context = {}
+        for yara_match in matches:
+            match_context = []
+            for string_match in yara_match.strings:
+                expression_keys = []
```
msm-cert, line 204: Since this is only used for presence checking, this should be a set instead of a list. But I think it's not necessary at all (see next comment)

src/tasks.py
```diff
@@ -140,6 +194,36 @@ def execute_yara(self, job: Job, files: List[str]) -> None:
...
+                    match_context.append(
+                        {
+                            "before": base64.b64encode(before).decode("utf-8"),
+                            "matching": base64.b64encode(matching).decode(
+                                "utf-8"
+                            ),
+                            "after": base64.b64encode(after).decode("utf-8"),
+                        }
+                    )
+                    context.update({str(yara_match): match_context})
```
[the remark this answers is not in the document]
msm-cert: Finally, this expression is overly complicated - I believe
```
context.update({A: B})
```
is equivalent to
```
context[A] = B
```

src/tasks.py
```diff
@@ -108,10 +154,18 @@ def execute_yara(self, job: Job, files: List[str]) -> None:
...
+
                 matches = rule.match(path)
                 if matches:
+                    data = self.read_file(path)
+                    context = self.get_match_context(data, matches)
+
                     self.update_metadata(
-                        job.id, orig_name, path, [r.rule for r in matches]
+                        job=job.id,
+                        orig_name=orig_name,
```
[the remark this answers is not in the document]
michalkrzem: I read somewhere that when we use such a notation, we don’t have to worry about the order of the arguments, and we already know during the function call what variables are being assigned to (if the naming differs). That was someone’s opinion; I’ll adapt to the project then. :)

src/tasks.py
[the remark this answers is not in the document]
michalkrzem: Here, in the match key, we have information about the string we are searching for. But okay, I’ll replace the list with a dictionry.

src/tasks.py
```diff
@@ -140,6 +157,28 @@ def execute_yara(self, job: Job, files: List[str]) -> None:
...
+    def get_match_context(
+        data: bytes, matches: List[yara.Match]
+    ) -> Dict[str, Dict[str, Dict[str, str]]]:
+        context = {}
+        for yara_match in matches:
+            match_context = {}
+            for string_match in yara_match.strings:
+                expression_key = string_match.instances[0]
+
+                (before, matching, after,) = read_bytes_with_context(
```
msm-cert, line 170: nitpick
```suggestion
                (before, matching, after,) = read_bytes_with_context(
```

src/tasks.py
```diff
@@ -290,3 +329,18 @@ def run_yara_batch(job_id: JobId, iterator: str, batch_size: int) -> None:
...
+    data: bytes, matched_length: int, offset: int, byte_range: int = 32
+) -> tuple[bytes, bytes, bytes]:
+    """Return `matched_length` bytes from `offset`, along with `byte_range` bytes before and after the match."""
+
+    before = data[max(0, offset - byte_range) : offset]
+    matching = data[offset : offset + matched_length]
+    after = data[
+        offset
+        + matched_length : min(len(data), offset + matched_length + byte_range)
+    ]
```
msm-cert, line 344: nitpick, but we save 2 LoC - only negative numbers are a problem so this works too
```suggestion
    before = data[max(0, offset - byte_range) : offset]
    matching = data[offset : offset + matched_length]
    after = data[offset + matched_length : offset + matched_length + byte_range]
```

msm-cert (approved):
Thanks! Looks good. Sorry for waiting

src/tasks.py
```diff
@@ -140,6 +157,28 @@ def execute_yara(self, job: Job, files: List[str]) -> None:
...
+        context = {}
+        for yara_match in matches:
+            match_context = {}
+            for string_match in yara_match.strings:
+                expression_key = string_match.instances[0]
+
+                (before, matching, after) = read_bytes_with_context(
+                    data, expression_key.matched_length, expression_key.offset
+                )
+                match_context[expression_key] = {
```
msm-cert, line 173: I'll fix the style in a second but... `expression_key` is an object. So match_context is not a `Dict[str, ...]`, so context is not `Dict[..., Dict[str, ...]]`, so something's wrong

msm-cert (changes requested):

msm-cert (approved):

PR #444, opened 2024-12-31
437 implement api to manage indexing queues

### LINKED ISSUE #437: Implement API to manage indexing queues
We plan to implement an automated file indexing in the future.

Right now (after #434), there is a database object representing indexed files, but there's no way to manipulate them - add files, remove them, etc.

In this issue I propose to add three endpoints to mquery:

### `POST /api/queue/{ursadb_id}`

Takes a JSON body with a list of file paths on the ursadb server to be indexed.

For example, POST to `/api/queue/default` with `["/mnt/samples/12345"] ` will tell ursadb with ID `default` that a file `/mnt/samples/12345` should be indexed.

This endpoint should support all QueuedFile fields, i.e. tags and index types, etc.

JSON body can look like this:

```
[
    { "path": "...", "types": [...], "tags": [...] },
    { "path": "...", "types": [...], "tags": [...] },
]
```
or:
```
{
    "paths": ["..."],
    "types": [...],
    "tags": [...] 
]
```
whichever is easier to implement.

### `GET /api/queue/{ursadb_id}`

Returns a JSON with enough information to show the queue status in the admin panel. Probably for now this should be just something like:

```
[
    {
        "ursadb_id": "default",
        "size": 1330
        # I can't think of anything more useful - maybe date of oldest and newest file?
    },
    ...
]
```

### `POST /api/queue/{ursadb_id}/clear`

Or just `DELETE /api/queue/{name}` if you're feeling RESTy - to be decided by implementer.
Takes no parameters, and removes all files from the specified queue (i.e. removes all `QueuedFile`s with the given ursadb_id from the  database).

Finally, all endpoints must be hidden behind appropriate permissions. We need two new user roles: `can_view_queues` and `can_manage_queues`.  I think every user should be able to see queue status, and admin should be able to manage them by default. Add the permissions to the groups by default (see `expand_role` function).

This should most likely be implemented more than one PR, to avoid one huge PR

michalkrzem:

We manually add files to the queue.

We have an API endpoint for adding files to QueuedFile in the database

**Test plan**
Endpoint api "api/queue/ursadb_id"

**Closing issues**

fixes #437


src/db.py
```diff
@@ -474,3 +477,15 @@ def alembic_upgrade(self) -> None:
...
+        with self.session() as session:
+            for file in file_paths:
+                obj = QueuedFile(
+                    ursadb_id=ursadb_id,
+                    path=file.path,
+                    index_types=file.index_types,
+                    tags=file.tags,
+                )
+            session.add(obj)
+            session.commit()
```
mickol34, line 491: Ostatnio, gdy próbowałem robić inne zadanie i miałem do stworzenia wiele elementów, opłacało się skorzystać z `bulk_insert_mappings` zamiast dodawać iteracyjnie obiekt po obiekcie: `https://stackoverflow.com/questions/34057006/using-bulk-insert-mappings`, ewentualnie jakakolwiek inna operacja do wstawiania wielu modeli na raz.

Plus wydaje mi się, że obecnie po prostu iteracyjnie nadpisujesz tylko zmienną `obj`, po czym dodajesz i commitujesz tylko ostatnią instancję.

michalkrzem: o kurka, faktycznie - jeszcze jedna tabulacja

src/app.py
```diff
@@ -551,7 +556,7 @@ def job_statuses(user: User = Depends(current_user)) -> JobsSchema:
 @app.delete(
     "/api/query/{job_id}",
     response_model=StatusSchema,
-    dependencies=[Depends(can_manage_queries)],
```
msm-cert, line 554: accidental change?

src/db.py
```diff
@@ -474,3 +477,17 @@ def alembic_upgrade(self) -> None:
         config_file = Path(__file__).parent / "alembic.ini"
         alembic_cfg = Config(str(config_file))
         command.upgrade(alembic_cfg, "head")
+
+    def set_queued_file(self, ursadb_id, file_paths):
```
msm-cert, line 481: `add_queued_file` maybe, or `enqueue_file`

msm-cert: (also ursadb: str, file_paths: list[str] guessing by variable names, but I didn't check the types)

msm-cert: or `List[FilePathInputSchema]` actually I guess

src/schema.py
```diff
@@ -105,3 +105,9 @@ class ServerSchema(BaseModel):
     openid_url: Optional[str]
     openid_client_id: Optional[str]
     about: str
+
+
+class FilePathInputSchema(BaseModel):
+    path: str
+    index_types: List[str]
```
msm-cert, line 112: We should validate the index_types somehow. I thought there's an enum for that, but apparently not. I think the easiest solution is to add a validator checking that every element is in `["gram3", "text4", "hash4", "wide8"]`

michalkrzem: Can i use List[Literal["gram3", "text4", "hash4", "wide8"]] from typing?

src/app.py
```diff
@@ -589,6 +596,69 @@ def serve_index(path: str) -> FileResponse:
...
+    )
+
+
+@app.delete(
+    "/api/queue/{ursadb_id}",
+    response_model=StatusSchema,
+    tags=["queue"],
+    dependencies=[Depends(can_manage_queues)],
+)
+def delete_queued_by_ursadb_id(ursadb_id: str):
```
msm-cert, line 650:
```suggestion
def delete_queue_by_id(ursadb_id: str):
```

src/app.py
```diff
@@ -589,6 +596,69 @@ def serve_index(path: str) -> FileResponse:
...
+    ursadb_exist = db.exist_ursadb(ursadb_id)
+    if ursadb_exist:
+        db.delete_queued_file(ursadb_id)
+        return StatusSchema(status="ok")
+
+    raise HTTPException(
+        status_code=400,
+        detail="Ursadb_id not found.",
+    )
```
msm-cert, line 659: I think this method can just do nothing if there are no files in a given queue (i.e. no error)

src/app.py
```diff
@@ -589,6 +596,69 @@ def serve_index(path: str) -> FileResponse:
...
+            ),
+            newest_file=EdgeOfFileSchema(
+                path=oldest_file.path,
+                created_at=newest_file.created_at,
+            ),
+        )
+    raise HTTPException(
+        status_code=400,
+        detail="Ursadb_id not found.",
+    )
```
msm-cert, line 641: I don't think this should raise an exception if there are no files in the queue. Unfortunately, this means that oldest_file and newest_file must be nullable (they won't have a value if a queue is empty)

src/app.py
```diff
@@ -589,6 +596,69 @@ def serve_index(path: str) -> FileResponse:
...
+    response_model=QueueStatusSchema,
+    tags=["queue"],
+    dependencies=[Depends(can_view_queues)],
+)
+def get_queue_status(ursadb_id: str):
+    queue_files = db.get_queued_files(ursadb_id)
+
+    if queue_files:
+        oldest_file = min(queue_files, key=lambda x: x.created_at)
+        newest_file = max(queue_files, key=lambda x: x.created_at)
```
msm-cert, line 624: If we don't use all the files, this should be done in the database (if there are many files this will be unnecessarily slow).

But we can discuss what's needed here

msm-cert (approved):
Looks really nice now, thanks

PR #445, opened 2025-01-13
Fix/keycloak timeout

### LINKED ISSUE #411: The default keycloak/openid logintimeout is very short
Investigate why in the default mquery configuration the login timeout is very short and needs constant reauthentication.

Fix that, such that either reauthentication is not needed (at most once a few weeks) or happens automatically.

Additionaly, when user gets logged out, if the instance allows anonymous access, the user should see things as anonymous user (instead of having to login - no way around it).

### LINKED ISSUE #425: Don't force relogin on expired token, when anonymous user has enough permissions.
Another issue related to the OIDC flows - right now, when:

* user logs in once
* then their token expires
* the user visits any page

The user will always be redirected to a login flow. Even if the page is accessible by an anonymous user. So this is not just logging the user out, this is always forcing the relogin.

When token expires, user should just be regularly logged out, and have permissions like any other anonymous user would.

MichalJura:
- [ x] I've read the [contributing guideline](https://github.com/CERT-Polska/mquery/blob/master/CONTRIBUTING.md).

Currently mquery does not revoke token by itself therefore forcing relogin.
Also after token expiration user is forced to log in even if instance allows anonymous access,

**Test plan**

Token is auto-refreshing itself now.
Also when the token expires, user get anonymous access.
Also User is not logged out so often.

**Closing issues**

fixes #411 
fixes #425

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -1,6 +1,8 @@
 from contextlib import asynccontextmanager
 import os
 
+import requests  # type: ignore
+
 import uvicorn  # type: ignore
 from pathlib import Path
 from fastapi import (
@@ -44,6 +46,9 @@ from .schema import (
     BackendStatusDatasetsSchema,
     AgentSchema,
     ServerSchema,
+    LoginSchema,
+    LogoutSchema,
+    RefreshTokenSchema,
 )
 
 
@@ -70,6 +75,22 @@ def with_plugins() -> Iterable[PluginManager]:
         plugins.cleanup()
 
 
+def get_new_token(refresh_token):
+    data = {
+        "grant_type": "refresh_token",
+        "refresh_token": refresh_token,
+        "client_id": db.config.openid_client_id,
+        "client_secret": db.config.openid_secret,
+    }
+    url = "http://mquery-keycloak-1:8080/auth/realms/myrealm/protocol/openid-connect/token"
+    try:
+        response = requests.post(url=url, data=data)
+        return response.json()
+
+    except requests.exceptions.RequestException:
+        return None
+
+
 class User:
     def __init__(self, token: Optional[Dict]) -> None:
         self.__token = token
@@ -92,6 +113,37 @@ class User:
             return [UserRole[name] for name in role_names]
         except KeyError:
             return []
+
+
+class TokenChecker:
+    def __init__(self):
+        self._tokens_data = {}
+
+    def show_tokens(self):
+        return self._tokens_data
+
+    def set_token(self, token, refresh_token):
+        self._tokens_data[token] = refresh_token
+
+    def refresh_token(self, token):
+        if token in self._tokens_data:
+            token_data = get_new_token(self._tokens_data[token])
+            if token_data:
+                new_token = token_data["access_token"]
+                new_refresh_token = token_data["refresh_token"]
+                del self._tokens_data[token]
+                self._tokens_data[new_token] = new_refresh_token
+                return new_token
+            return None
+        else:
+            return None
+
+    def remove_token(self, token):
+        if token in self._tokens_data:
+            del self._tokens_data[token]
+
+
+token_checker = TokenChecker()
 
 
 async def current_user(authorization: Optional[str] = Header(None)) -> User:
@@ -125,7 +177,7 @@ async def current_user(authorization: Optional[str] = Header(None)) -> User:
             token, public_key, algorithms=["RS256"], audience="account"  # type: ignore
         )
     except jwt.ExpiredSignatureError:
-        # token expired, so user is now anonymous
+        # token expired so user is anonymous
         return User(None)
     except jwt.InvalidTokenError:
         # Invalid token means invalid signature, issuer, or just expired.
@@ -303,6 +355,7 @@ def backend_status_datasets() -> BackendStatusDatasetsSchema:
 
     This endpoint is not stable and may be subject to change in the future.
     """
+    logging.error(f"NEW TOKEN\n{token_checker.show_tokens()}\n***** ")
     datasets: Dict[str, int] = {}
     for agent_spec in db.get_active_agents().values():
         try:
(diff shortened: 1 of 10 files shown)


src/app.py
```diff
@@ -70,6 +74,24 @@ def with_plugins() -> Iterable[PluginManager]:
...
 
 
+def get_new_tokens(refresh_token):
+    data = {
+        "grant_type": "refresh_token",
+        "refresh_token": refresh_token,
+        "client_id": db.config.openid_client_id,
+        "client_secret": db.config.openid_secret,
+    }
+    url = "http://mquery-keycloak-1:8080/auth/realms/myrealm/protocol/openid-connect/token"
```
msm-cert, line 84: Shouldn't be hardcoded - use url from config instead

src/app.py
```diff
@@ -70,6 +74,24 @@ def with_plugins() -> Iterable[PluginManager]:
         plugins.cleanup()
 
 
+def get_new_tokens(refresh_token):
```
msm-cert, line 77: Please add typing (I guess `get_new_tokens(refresh_token: str) -> Tuple[str, str]`, but recheck)

src/app.py
```diff
@@ -1,6 +1,8 @@
 from contextlib import asynccontextmanager
 import os
 
+import requests  # type: ignore
+
```
msm-cert, line 5:
```suggestion
```
(I accidentaly removed this comment now, but it still applies - unnecessary empty line.

msm-cert (commented):
Ping about hardcoded URL and style nitpicks.

I feel verifying OIDC correctness here is out of my depth, so I asked @psrok1 to help with the review of this aspect.

src/mqueryfront/src/utils.js
```diff
@@ -25,3 +27,45 @@ export const openidLoginUrl = (config) => {
...
+    clearInterval(tokenInterval);
+    localStorage.removeItem("expiresAt");
+    localStorage.removeItem("rawToken");
+};
+
+export const tokenExpired = () => {
+    const rawToken = localStorage.getItem("rawToken");
+    if (rawToken) {
+        const expiresAt = localStorage.getItem("expiresAt");
+        if (Date.now() > expiresAt) {
```
msm-cert, line 65: Probably it would be better to check if token expires in the next ~30 seconds? To avoid situations where token is 0.1 second before expiration during the check, but server gets an expired token already. That, and clock desync.

src/mqueryfront/src/utils.js
```diff
@@ -25,3 +27,45 @@ export const openidLoginUrl = (config) => {
...
+
+export const clearTokenData = (tokenInterval) => {
+    clearInterval(tokenInterval);
+    localStorage.removeItem("expiresAt");
+    localStorage.removeItem("rawToken");
+};
+
+export const tokenExpired = () => {
+    const rawToken = localStorage.getItem("rawToken");
+    if (rawToken) {
```
msm-cert, line 63: nitpick - if possible prefer early exit and avoid nested ifs. So please change the flow to

```
if (!rawToken) {
    return false;
}
```

src/mqueryfront/src/utils.js
```diff
@@ -25,3 +27,45 @@ export const openidLoginUrl = (config) => {
...
+        const headers = rawToken ? { Authorization: `Bearer ${rawToken}` } : {};
+        const response = await axios.request("/api/token/refresh", {
+            method: "POST",
+            headers: headers,
+            withCredentials: true,
+        });
+        if (response.data["token"]) {
+            storeTokenData(response.data["token"]);
+        } else {
+            return;
```
msm-cert, line 50: Does this return here do anything? If not, please remove it.

src/mqueryfront/src/utils.js
```diff
@@ -25,3 +27,45 @@ export const openidLoginUrl = (config) => {
...
+
+export const refreshAccesToken = async () => {
+    const rawToken = localStorage.getItem("rawToken");
+    const expiresAt = localStorage.getItem("expiresAt");
+    if (rawToken) {
+        const headers = rawToken ? { Authorization: `Bearer ${rawToken}` } : {};
+        const response = await axios.request("/api/token/refresh", {
+            method: "POST",
+            headers: headers,
+            withCredentials: true,
```
msm-cert, line 45: Why is `withCredentials` here necessary? If it's not needed, please remove it.

src/mqueryfront/src/App.js
```diff
@@ -21,20 +22,32 @@ function getCurrentTokenOrNull() {
...
+    const login = async (token_data) => {
+        token_data.not_before_policy = token_data["not-before-policy"];
+        delete token_data["not-before-policy"];
+        const response = await api.post("/login", token_data);
+        storeTokenData(token_data["access_token"]);
+        const location_href = localStorage.getItem("currentLocation");
+        if (location_href) {
+            window.location.href = location_href;
+        } else {
+            window.location.href = "/";
```
msm-cert, line 45: Nitpick: this is technically incorrect, since mquery doesn't have to be hosted at `/` (it won't work with, for example, a web proxy that serves mquery on `some.website.pl/mquery/`). But I think this problem is safe to ignore, just wanted to point that out.

src/mqueryfront/src/App.js
```diff
@@ -21,20 +22,32 @@ function getCurrentTokenOrNull() {
...
+        }, 900000); // refresh token every 15 minutes just in case user was idle.
+        return () => clearInterval(tokenIntervalRef.current);
     }, []);
-
-    const login = (rawToken) => {
-        localStorage.setItem("rawToken", rawToken);
-        window.location.href = "/";
+    const login = async (token_data) => {
+        token_data.not_before_policy = token_data["not-before-policy"];
+        delete token_data["not-before-policy"];
```
msm-cert, line 38: Are these steps necessary? I don't know much about OIDC but it feels weird that we have to do this.

src/mqueryfront/src/App.js
```diff
@@ -21,20 +22,32 @@ function getCurrentTokenOrNull() {
...
+        const response = await api.post("/login", token_data);
```
msm-cert, line 39:
```suggestion
        await api.post("/login", token_data);
```
since response is not used anywhere

src/mqueryfront/src/config/ConfigPage.js
```diff
@@ -14,6 +14,7 @@ class ConfigPage extends Component {
     }
 
     componentDidMount() {
+        localStorage.setItem("currentLocation", window.location.href);
```
msm-cert, line 17: Are you sure there is no way to solve this in a more generic way? Adding the code to set a variable on every subpage feels suboptimal and error-prone.

src/mqueryfront/src/utils.js
```diff
@@ -25,3 +27,45 @@ export const openidLoginUrl = (config) => {
     );
     return login_url;
 };
+
+export const storeTokenData = (token) => {
+    localStorage.setItem("rawToken", token);
+    const decodedToken = parseJWT(token);
+    localStorage.setItem("expiresAt", decodedToken.exp * 1000);
```
msm-cert, line 34: Please don't keep this in localStorage. It's better to avoid having two separate variables when one is necessary. Instead parse rawToken when necessary.

src/mqueryfront/src/utils.js
```diff
@@ -25,3 +27,45 @@ export const openidLoginUrl = (config) => {
...
+};
+
+export const refreshAccesToken = async () => {
+    const rawToken = localStorage.getItem("rawToken");
+    const expiresAt = localStorage.getItem("expiresAt");
```
msm-cert, line 39: unused

src/mqueryfront/src/utils.js
msm-cert: Nevermind, I just noticed the code uses cookies.

src/app.py
```diff
@@ -584,6 +609,35 @@ def server() -> ServerSchema:
     )
 
 
+@app.post("/api/login", response_model=LoginSchema, tags=["stable"])
+async def login(request: Request, response: Response) -> LoginSchema:
+    token = await request.json()
+    if token["refresh_token"]:
```
msm-cert, line 615: The code intentionally doesn't use cookies and passes all parameters by the API. Please change this to keep refresh_token in local storage instead of adding a cookie.

msm-cert (changes requested):
I did a more in-depth review and found a few more stylistic problems. Also a few design comments (especially cookie usage).

msm-cert:
(also please rebase this onto master, since there are merge conflicts)

PR #448, opened 2025-02-27
fix: refresh ursadb page and add alert about successful submit

mickol34:

Currently theres no feedback to user about successful submit to UrsaDB page.

This PR introduces new alert about successful POST request and clears form upon submission

**Test plan**

Click through new page and check for any unexpected behavior

**Closing issues**

fixes #issuenumber

### DIFF
--- a/src/mqueryfront/src/App.css
+++ b/src/mqueryfront/src/App.css
@@ -95,6 +95,14 @@
     right: 5vw;
 }
 
+.modal-container-index-page {
+    position: fixed;
+    top: 50%;
+    left: 50%;
+    transform: translate(-50%, -50%);
+    z-index: auto;
+}
+
 .modal-block {
     position: relative;
     block-size: "fit-content";
--- a/src/mqueryfront/src/App.css
+++ b/src/mqueryfront/src/App.css
@@ -100,7 +100,6 @@
     top: 50%;
     left: 50%;
     transform: translate(-50%, -50%);
-    z-index: auto;
 }
 
 .modal-block {
--- a/src/mqueryfront/src/indexFiles/IndexPage.js
+++ b/src/mqueryfront/src/indexFiles/IndexPage.js
@@ -268,10 +268,14 @@ class IndexPageInner extends Component {
                     </div>
                     <div className="my-2">{`Files in queue (regardless of status): ${this.state.size}`}</div>
                     {this.state.newestFile && (
-                        <div className="my-2">{`Newest file in the queue: ${this.state.newestFile.replace('T', ' ').slice(0, -7)}`}</div>
+                        <div className="my-2">{`Newest file in the queue: ${this.state.newestFile
+                            .replace("T", " ")
+                            .slice(0, -7)}`}</div>
                     )}
                     {this.state.oldestFile && (
-                        <div className="my-2">{`Oldest file in the queue: ${this.state.oldestFile.replace('T', ' ').slice(0, -7)}`}</div>
+                        <div className="my-2">{`Oldest file in the queue: ${this.state.oldestFile
+                            .replace("T", " ")
+                            .slice(0, -7)}`}</div>
                     )}
                     <IndexClearQueueButton
                         msg="Clear queue"
--- a/src/mqueryfront/src/indexFiles/IndexPage.js
+++ b/src/mqueryfront/src/indexFiles/IndexPage.js
@@ -5,6 +5,8 @@ import api from "../api";
 import ErrorBoundary from "../components/ErrorBoundary";
 import IndexClearQueueButton from "./IndexClearQueueButton";
 import IndexSuccessPage from "./IndexSuccessPage";
+import IndexClearedPage from "./IndexClearedPage";
+import Draggable from "react-draggable";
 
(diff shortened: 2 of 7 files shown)


src/mqueryfront/src/indexFiles/IndexPage.js
```diff
@@ -72,6 +83,10 @@ class IndexPageInner extends Component {
         this.setState({ selectedTaints: selection });
     }
 
+    fileOrFiles(length) {
+        return `file${length > 1 ? "s" : ""}`;
+    }
```
msm-cert, line 88: Nice, I wondered about suggesting this refactoring in the previous PR.

But it's usually handled a bit differently, with a helper like `pluralize`:
```
const pluralize = (count, noun) =>
  `${count} ${noun}${count !== 1 ? "s" : ''}`;
```

This is a bit more generouns, as you can use it for things other than file. More importantly, you don't have to do 

`${this.state.alertShowFileLen} ${this.fileOrFiles(this.state.alertShowFileLen)}!`

and instead can just

`${pluralize("files", this.state.alertShowFileLen}}!`

which is nice I guess

src/mqueryfront/src/indexFiles/IndexPage.js
```diff
@@ -81,17 +96,138 @@ class IndexPageInner extends Component {
...
+                                this.state.alertShowFileLen
+                            } ${this.fileOrFiles(
+                                this.state.alertShowFileLen
+                            )}!`}
+                            onClick={this.handleAlertClose}
+                        />
+                    )}
+                    {this.state.alertShowCleared && (
+                        <IndexClearedPage
+                            msg={`Successfully cleared ${
```
msm-cert, line 223: `Successfully cleared false file from queue default!`

I think the variable here should be `alertShowCleared`?

src/mqueryfront/src/indexFiles/IndexPage.js
```diff
@@ -104,23 +240,26 @@ class IndexPageInner extends Component {
...
                         >
                             {`Add to queue${
                                 fileLen > 0
-                                    ? ` (${fileLen} file${
-                                          fileLen > 1 ? "s" : ""
-                                      })`
+                                    ? ` (${fileLen} ${this.fileOrFiles(
+                                          fileLen
+                                      )})`
                                     : ""
```
msm-cert, line 263: The same comment about pluralize here - it will work if you reorder it to `add N files to queue` which also makes it a bit more idiomatic english

src/mqueryfront/src/App.css
```diff
@@ -95,6 +95,14 @@
     right: 5vw;
 }
 
+.modal-container-index-page {
+    position: fixed;
+    top: 50%;
+    left: 50%;
+    transform: translate(-50%, -50%);
+    z-index: auto;
```
msm-cert, line 103: isn't `z-index: auto` the default value? Unless there's some inheritance going on here.

src/mqueryfront/src/indexFiles/IndexPage.js
```diff
@@ -130,11 +269,14 @@ class IndexPageInner extends Component {
                         <div className="my-2">{`Newest file in the queue: ${this.state.newestFile}`}</div>
```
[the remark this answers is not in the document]
mickol34: Should I also remove seconds or maybe they are still relevant? For now maybe I'll leave seconds, later this can easily be changed by changing slicing index.

src/mqueryfront/src/App.css
mickol34: Leftover from modal styling :grin: Will clear this.

src/mqueryfront/src/indexFiles/IndexPage.js
mickol34: Yeah, I'll generalize this function to all nouns, but I think it would be better to leave first ${count} out of this function.

src/mqueryfront/src/indexFiles/IndexPage.js
msm-cert: step one: add 1 file
image
step two: add 2 files
image
now there are three files in the queue (verify by `select * from queuedfile;`)
step three: clear queue
this should clear three fules, but the result is either:
image
or
image
depending on if the previous banner was closed. Also in the latter situation, I can now add one file again, and get:
image
so I believe there is some variable confusion here :duck:

You're right that my comment was wrong, I think that the issue is that `handleClearQueue` doesn't set FileLen? But I leave this to you, anyway something's wrong I belive

msm-cert (changes requested):

src/mqueryfront/src/indexFiles/IndexPage.js
mickol34: Yes, I mistakenly used the same variable used to store sent files to also show amount of cleared files. Unfortunately I think the best way to handle this issue was to simplify cleared queue alert not to use any numbers, since it's non-obvious how many files are currently in queue after submitting some amount (and later if queued files would happen to change status server-side whilst page would not update).

msm-cert (approved):
I tested it and looks good!

PR #459, opened 2026-03-06
Add an optional link between mquery results and mwdb

dev-cti:
- [X ] I've read the [contributing guideline](https://github.com/CERT-Polska/mquery/blob/master/CONTRIBUTING.md).
- [ X] I've tested my changes by building and running mquery, and testing changed functionality (if applicable)

There is currently no way to switch between mquery results and mwdb 
The user has to copy the sample hash and select it in mwdb

A new option is added to the global Mquery plugin to be able to specify the mwdb instance url.
If this option is set, links are displayed in the mquery results panels. They can be used to switch to mwdb easily.
If this options is not set, the current behaviour is unchanged

**Test plan**
- Test 1

> 1. Configure the key mwdb_url of Mquery plugin  in the config panel (http://localhost:config) with the MWDB base url
> 2. Check that yara results are now displayed as links
> 3. Check that a new tab with the sample MWDB page is opened on click 

- Test 2

> 1. Remove the key mwdb_url of Mquery plugin  in the config panel (http://localhost:config) with the MWDB base url
> 2. Check that yara results are not displayed as links

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -580,6 +580,7 @@ def server() -> ServerSchema:
         auth_enabled=str(db.config.auth_enabled).lower(),
         openid_url=db.config.openid_url,
         openid_client_id=db.config.openid_client_id,
+        mwdb_url=db.config.mwdb_url,
         about=app_config.mquery.about,
     )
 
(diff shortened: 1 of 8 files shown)


msm-code:
Hi,
thanks for your PR. I'll review it shortly.

>There is currently no way to switch between mquery results and mwdb

We used to have an internal plugin that worked around that. But that was not very polished and relied on a small hack that is now not necessary. I think that the idea behind this PR is good - thanks for sending again and I'll get back to you soon.

